PR #9, opened 2021-02-25
Change `ditto-cli` to `ditto`

jolenam:
## Overview
Update commands (and README) with `ditto` instead of `ditto-cli`


ouyangatang (approved):
lgtm

PR #16, opened 2021-08-18
[DIT-271] Fix error when running `project add` on full YAML file

azjgard:
## Overview

The CLI was incorrectly returning a status code of 2 (signaling that something went wrong with the execution of the program) when a user attempts to add a project to a YAML file already containing all of the projects in the workspace.

This PR fixes that to return a status code of 0 in that instance; status code 2 should only be used when the program encounters an error causing it to exit early.

## Context

Linear ticket:
https://linear.app/dittowords/issue/DIT-271/fix-ditto-cli-error-from-adding-project-to-yaml-file

## Screenshots
Before:
image
image

After:
image
image

## Test Plan
- [ ] Add all projects in workspace to YAML file
- [ ] Run `npx ditto-cli project add`
- [ ] Confirm no error

For some reason I couldn't get the error to throw by running the CLI directly (note that in Jason's original report it was NPM itself that was throwing the error based on the status code); I instead created a test project with this in `package.json`:
```json
  "scripts": {
    "prepare": "DITTO_API_HOST=http://localhost:3001 npx ditto-cli project"
  }
  ```
  
  Then running `yarn install` or `npm install` reproduces the error. To test the local branch, you can replace the above with:
  ```json
  "scripts": {
    "prepare": "DITTO_API_HOST=http://localhost:3001 node ~/projects/ditto-cli/bin/ditto.js"
  }
  ```
  
You obviously don't need `DITTO_API_HOST` set if you're testing against production, although in that case it'll be a little tricker to get ALL of the workspace projects into the yaml file

### DIFF
--- a/bin/ditto.js
+++ b/bin/ditto.js
@@ -15,9 +15,9 @@ const removeProject = require('../lib/remove-project');
  * @param {any} error The thrown error object.
  * @returns {void}
  */
-function quit() {
+function quit(exitCode = 2) {
   console.log('\nExiting Ditto CLI...\n');
-  process.exitCode = 2;
+  process.exitCode = exitCode;
   process.exit();
 }
 
--- a/lib/add-project.js
+++ b/lib/add-project.js
@@ -2,9 +2,9 @@ const { collectAndSaveProject } = require('./init/project');
 const projectsToText = require('./utils/projectsToText');
 const getSelectedProjects = require('./utils/getSelectedProjects');
 
-function quit() {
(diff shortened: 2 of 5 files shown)


ouyangatang (approved):
Looks good to me!! 🥳 Thank you so much for the quick fix + responding to Jason's messages in the channel

One other thing before we re-publish the CLI -- could you up the version in the package.json (to 1.0.1 as it's a bug fix)? I believe it needs to be different version in order for it to be accepted by npm.

PR #17, opened 2021-09-21
[DIT-360] Add optional support for variants and non-default formats

azjgard:
## Overview

- Adds a precursor action to the `pull` command that cleans the output directory of all `.json` files prior to fetching new data
- Adds support for two new options in `config.yml`:
1. `variants` 
>- possible values: `true | false`
>- if `true`, variant text will be fetched for specified projects, and a separate file will be written for each variant
>- if `true`, the base text for specified projects will be written to `base.json` instead of `text.json`
2. `format`
>- possible values: `flat | structured`
>- if specified, all text (regardless of whether variants are specified) will be fetched and stored in the indicated format

## Context

https://linear.app/dittowords/issue/DIT-360/cli-changes-flat-json-and-variants

## Screenshots
image

## Test Plan
- [ ] `text.json` is generated in the default format when no new options are configured
- [ ] `base.json` and variant files are generated when `variants` option is set to `true`
- [ ] Files are generated in `flat` format when `format` is set to `flat`
- [ ] Files are generated in `structured` format when `format` is set to `structured`
- [ ] `base.json` and variant files are generated in specified format when `variants` is set to `true` and a non-default `format` is specified

## TODO Post-Review
- [ ] Change version from `1.1.1-beta` to `1.1.1` and re-publish to NPM


XavierAgostini (approved):
just tested each option and it worked great for me!

PR #21, opened 2021-10-28
[DIT-571] Only write variant files for projects that have them

azjgard:
## Overview
Prior to this PR, files would always be written for every variant in the workspace if `variants: true` was specified. This PR changes that behavior so that variant files are only written for projects that actually contain values for a given variant.

https://github.com/dittowords/ditto-app/pull/687 is a companion PR to this one, but its functionality is not required for this PR to be tested - it only reduces the number of requests that need to be made.

## Context
https://linear.app/dittowords/issue/DIT-571/cli-only-output-variant-files-for-variants-with-values-in-specified

## Screenshots
before:
image

after:
image

## Test Plan
-  [x] Pull with `variants: true` and confirm that no blank files are generated

### DIFF
--- a/lib/pull.js
+++ b/lib/pull.js
@@ -88,6 +88,11 @@ async function downloadAndSaveVariant(variantApiId, projects, format, token) {
           headers: { Authorization: `token ${token}` },
         });
 
+        // only write files that have data
+        if (!Object.keys(data).length) {
+          return "";
+        }
+
         const filename = name + ("__" + (variantApiId || "base")) + ".json";
         const filepath = path.join(consts.TEXT_DIR, filename);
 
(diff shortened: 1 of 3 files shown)


lib/pull.js
```diff
@@ -298,7 +336,11 @@ async function downloadAndSave(projectConfig, token) {
   spinner.start();
 
   try {
-    msg += cleanOutputFiles();
+    msg += await cleanOutputFiles();
+
+    // terrible stopgap solution to avoid some sort of unseen
+    // race condition
+    await new Promise((resolve) => setTimeout(resolve, 3000));
```
ouyangatang, line 343: do we want to remove this before merging?

ouyangatang (approved):
This looks good to me! The only thing is updating the version back down to 1.2.4 (and to remove the stopgap solution) -- feel free to merge afterwards. 

Thank you so much for making these changes!! And sorry for the headache-inducing testing that I was embroiled in last week 😅

PR #25, opened 2022-01-19
Support passing API token via environment variable

azjgard:
## Overview
Support passing an API token via an environment variable `DITTO_API_KEY`.

If environment variable is set, it will be used with precedence over any tokens stored in `~/.config/ditto`.

## Context
Interacting with the Ditto CLI from CI environments or deployed applications will be made way easier by allowing the environment variable as an alternative to the current config file approach.

## Test Plan
- [ ] Test all commands with the existing approach, where the API token is stored in `~/.config/ditto`:
>- [ ] no command (should just display help)
>- [ ] `project`
>- [ ] `project add`
>- [ ] `project remove`
>- [ ] `pull`
- [ ] Test all commands with the the new approach by deleting `~/.config/ditto` and passing the API token in `DITTO_API_KEY` (commands should work without being prompted for a token):
>- [ ] no command (should just display help)
>- [ ] `project`
>- [ ] `project add`
>- [ ] `project remove`
>- [ ] `pull`

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,17 @@ After selecting a project, a configuration file will automatically be created at
 
 Once you've successfully authenticated and a config file has been created, you’re ready to start fetching copy! You can set up the CLI in multiple directories by running `ditto-cli` and choosing an initial project to sync from.
 
+## API Keys
+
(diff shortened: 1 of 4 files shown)


asnewman (approved):
Excellent work! Worked locally for me 🚀

PR #26, opened 2022-02-24
Allow users to override the output directory

acao:
## Overview

Allows users to override the text output directory using DITTO_TEXT_DIR env variable

## Context

often with typescript/etc, modules cannot resolve paths at the repo root or outside of the `src` or equivalent directory

### DIFF
--- a/lib/consts.js
+++ b/lib/consts.js
@@ -1,6 +1,7 @@
 const homedir = require("os").homedir();
-
 const path = require("path");
+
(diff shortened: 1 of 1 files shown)


azjgard:
Hi @acao, thanks so much for opening this! 

The change looks good to me ✨ I'll merge and include this in our next release.

azjgard (approved):

PR #32, opened 2022-05-19
Add meta flag

asnewman:
## Overview
Adds a `-m, --meta <data...>` flag that takes in keypairs delimited by `:`s and separated by ` `s (ex: `-m githubActionRequest:true trigger:manual`

## Context
This will be used to signal the Ditto backend when the CLI is running within a GitHub action. Because this option is generic, the hope is that it will be used in the future to send any arbitrary data to the Ditto backend.

## Test Plan

- [x] Verify the option shows up when running `ditto-cli`
- [x] Verify meta options are being parsed correctly
- [x] Tests pass


lib/utils/processMetaOption.test.js
```diff
@@ -0,0 +1,16 @@
...
+      trigger: 'manual',
+    });
+  });
+  it('Malformed doesnt crash', () => {
+    expect(processMetaOption(['context:github-action', 'trigger'])).toEqual({
+      context: 'github-action',
+      trigger: undefined,
+    });
+  });
+});
```
azjgard, line 16: 👍

lib/pull.js
```diff
@@ -334,11 +351,19 @@ async function downloadAndSave(sourceInformation, token) {
...
+/**
+ * @param {{ meta: Object.<string, string> }} options
+ */
+function pull(options) {
+  const { meta } = options;
   const token = config.getToken(consts.CONFIG_FILE, consts.API_HOST);
   const sourceInformation = config.parseSourceInformation();
 
-  return downloadAndSave(sourceInformation, token);
+  if (meta?.context === 'github-action') {
```
azjgard, line 362: In the near future, I want to.. 
- convert this entire repository to TypeScript
- pin the minimum required Node version

Until then though, I think we should avoid using optional chaining (and other features not introduced till later versions) since this could technically cause breakage for anyone using our CLI with an older version of Node.

Let me know if you disagree!

lib/pull.js
```diff
@@ -107,8 +114,15 @@ async function downloadAndSaveVariants(projects, format, token) {
   return messages.join("");
 }
 
-async function downloadAndSaveBase(projects, format, token) {
-  const params = {};
+/**
+ * @param {{ githubActionRequest: boolean }} options
+ */
+async function downloadAndSaveBase(projects, format, token, options) {
+  const githubActionRequest = options?.githubActionRequest || false;
```
azjgard, line 121: Same as https://github.com/dittowords/cli/pull/32/files#r879741599

azjgard:
DM'ed a suggestion about making the `meta` config flag truly generic and flexible for the future by avoiding adding coupled GH Action code to this repository.

azjgard (approved):
Looks great!

PR #33, opened 2022-06-27
Jrod/dit 1695 fix ditto cli project add command when

jaerod95:
## Overview

Fixes adding and removing the component library from the cli config

## Screenshots

Screen Shot 2022-06-27 at 9 24 10 AM

## Test Plan

Testing successfully completed in local env via:

- [ ] add the component library and also remove it using the `project add` and `project remove` commands (should set the components field to true or false and not modify the project list
- [ ] make sure adding and removing normal projects work and update the config.yml. The wording should change depending if you've added the component library & projects vs just projects vs just component library.


lib/utils/getSelectedProjects.ts
```diff
@@ -0,0 +1,67 @@
...
+    return [];
+  }
+
+  return contentJson.projects.filter(({ name, id }) => name && id);
+};
+
+/**
+ * Returns an array containing all valid projects ({ id, name })
+ * currently contained in the project config file.
+ */
```
azjgard, line 59: I think this comment was copied and pasted and isn't relevant for this function.

lib/remove-project.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  const allProjects = isUsingComponents
+    ? [{ id: "components", name: "Ditto Component Library" }, ...projects]
+    : projects;
+
+  const projectToRemove = await promptForProject({
+    projects: allProjects,
+    message: isUsingComponents
+      ? "Select a project or library to remove"
```
azjgard, line 29: I wonder if we should update our nomenclature in the CLI to refer to "sources" instead of "projects", since places like this are a little bit awkward. "Select a project or library to remove" is a little confusing bc it implies that a "library" is a thing, where really we're only talking about THE component library.

If we were to switch over to saying internally "a project or the component library are a type of 'source' the CLI uses", then this message could say "Select a source to remove" and be less ambiguous.

Something for us to noodle on 🤔

lib/init/project.js
```diff
@@ -18,7 +21,7 @@ function saveProject(file, name, id) {
   // old functionality included "ditto_component_library" in the `projects`
   // array, but we want to always treat the component library as a separate
   // entity and use the new notation of a top-level `components` key
-  if (id === "ditto_component_library") {
+  if (id === "components") {
```
azjgard, line 24: Looks like the comment here is slightly out of date now

azjgard (changes requested):
It looks like things currently break if you run `project add` or `pull` and a config file doesn't currently exist:
image

azjgard:
Offline conversation: we're going to break this into a couple of PRs.

- A standalone PR with the bugfix that we can quickly test and get out
- A standalone PR with the TypeScript changes / refactors that we can take our time testing and roll out in due time

PR #34, opened 2022-06-28
Fixes bug adding and removing ditto component library

jaerod95:
## Overview

This PR fixes a bug where adding and removing the Ditto Component Library wouldn't actually modify the config.yml file.

## Screenshots

image

## Test Plan

Testing successfully completed in local env via:

- [x] add the component library and also remove it using the `project add` and `project remove` commands (should set the components field to true or false and not modify the project list
- [x] make sure adding and removing normal projects work and update the config.yml. The wording should change depending if you've added the component library & projects vs just projects vs just component library.

### DIFF
--- a/lib/add-project.js
+++ b/lib/add-project.js
@@ -1,6 +1,10 @@
 const { collectAndSaveProject } = require("./init/project");
 const projectsToText = require("./utils/projectsToText");
-const getSelectedProjects = require("./utils/getSelectedProjects");
+const {
+  getSelectedProjects,
+  getIsUsingComponents,
+} = require("./utils/getSelectedProjects");
+const output = require("./output");
 
 function quit(exitCode = 2) {
   console.log("Project selection was not updated.");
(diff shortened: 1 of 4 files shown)


lib/remove-project.js
```diff
@@ -15,13 +19,20 @@ async function removeProject() {
     return;
   }
 
+  const allProjects = isUsingComponents
+    ? [{ id: "components", name: "Ditto Component Library" }, ...projects]
```
azjgard, line 23: On the backend of the public API, we still have some legacy code in project routes that treat the component library as a "special project" according to the identifier "ditto_component_library"; I think for consistency's sake it could be good to continue using that here, and hopefully eventually we can just completely move away from the pattern.

lib/add-project.js
```diff
@@ -10,13 +14,30 @@ function quit(exitCode = 2) {
...
+    if (usingComponents) {
+      if (projects.length) {
+        console.log(
+          `\nYou're currently syncing text from the ${output.info(
+            "Component Library"
+          )} and from the following projects: ${projectsToText(projects)}`
+        );
+      } else {
+        console.log(
+          `\nYou're currently only syncing text from the ${output.info(
```
azjgard, line 29:
```suggestion
          `\nYou're currently set up to sync text from the ${output.info(
```

lib/add-project.js
```diff
@@ -10,13 +14,30 @@ function quit(exitCode = 2) {
...
   try {
-    console.log(
-      `\nYou're currently set up to sync text from the following projects: ${projectsToText(
-        projects
-      )}`
-    );
+    if (usingComponents) {
+      if (projects.length) {
+        console.log(
+          `\nYou're currently syncing text from the ${output.info(
```
azjgard, line 23:
```suggestion
          `\nYou're currently set up to sync text from the ${output.info(
```

azjgard (changes requested):
Functionality looks great 👍 left a few very minor comments, once those are addressed will briefly re-test and then we can 🚢

lib/remove-project.js
jaerod95: sounds good pushed up a fix

azjgard (approved):

PR #35, opened 2022-06-28
Jrod/update to typescript

jaerod95:
## Overview

Adds Typescript support for the cli project and migrates all js files to typescript.

## Context

Typescript will reduce the risk of bugs going forward with type safety and generally improve the developer experience of the cli project.

## Test Plan

Testing will require a thorough regression of the entire cli project and test suite.
[] Test Everything!


azjgard (approved):
Tested the following:
- Project 
- Project add
- Project remove
- No command with config file
- No command without a config file
- With / without configuration file
- With / without a stored auth token
- Pull
    - With default format variants
    - With default format no variants
    - With structured format variants
    - With structured format no variants
    - With flat format variants
    - With flat format no variants
    - With components: true
    - With components: false

Everything looks good 👍

lib/init/project.ts
```diff
@@ -43,17 +45,17 @@ async function askForAnotherToken() {
...
-  componentsSelected
+  token: Token,
+  projectsAlreadySelected: Project[],
+  componentsSelected: boolean
 ) {
   const spinner = ora("Fetching projects in your workspace...");
   spinner.start();
 
-  let projects = [];
+  let response: AxiosResponse<any>;
```
azjgard, line 55: I think this `any` could be replaced with `{ id: string; name: string; }[]`

lib/init/token.ts
```diff
@@ -1,38 +1,38 @@
...
   )
     return true;
   return false;
-}
+};
 
 // Returns true if valid, otherwise an error message.
-async function checkToken(token) {
-  const axios = api.create(token);
+async function checkToken(token: string): Promise<any> {
```
azjgard, line 29: I think you can remove this `any` and ts will infer the return type

lib/pull.ts
```diff
@@ -1,20 +1,21 @@
...
+import { collectAndSaveToken } from "./init/token";
+import sourcesToText from "./utils/sourcesToText";
+import { SourceInformation, Token, Project } from "./types";
 
 const NON_DEFAULT_FORMATS = ["flat", "structured"];
 
 const DEFAULT_FORMAT_KEYS = ["projects", "exported_at"];
-const hasVariantData = (data) => {
+const hasVariantData = (data: any) => {
+  // TODO: Type this to a projects response interface
```
azjgard, line 18: todo still relevant?

lib/pull.ts
```diff
@@ -314,7 +341,7 @@ async function downloadAndSave(sourceInformation, token, options) {
 
     spinner.stop();
     return console.log(msg);
-  } catch (e) {
+  } catch (e: any) {
```
azjgard, line 344:
```suggestion
  } catch (e: Error) {
```

lib/init/project.ts
```diff
@@ -134,7 +135,7 @@ async function collectAndSaveProject(initialize = false) {
     );
 
     saveProject(consts.PROJECT_CONFIG_FILE, project.name, project.id);
-  } catch (e) {
+  } catch (e: any) {
```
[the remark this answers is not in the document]
jaerod95: this results in the typing error

PR #41, opened 2022-11-30
Add status filtering

asnewman:
## Overview
This pull request adds a new optional config value called `status` which will filter text items/components when running `ditto pull`.

## Test Plan
Create a workspace with the following:
- Workspace components with variants and various statuses
- A project with text items that have variants and various statuses

Verify expected values with the following config values:
- [ ] components: true, variants: true, status: FINAL
- [ ] components: true, variants: false, status: FINAL
- [ ] components: true, variants: false
- [ ] components: true, variants: true

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,16 @@ If you run the CLI in a directory that does not contain a `ditto/` folder, the f
   format: flat
   ```
 
+  ##### `status`
+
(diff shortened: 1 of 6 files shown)


XavierAgostini (approved):
filter worked for me

PR #46, opened 2023-02-03
[DIT-3241] v3.0.0: general refactoring, update defaults, component folders

azjgard:
### Changes
- General refactoring & cleanup
#### Breaking changes
- New defaults:
    - `components: true`
    - `format: flat`
    - `variants: true`
- `projects` and `components` keys should now be nested under a top-level `sources` key
- `ditto_component_library` is no longer supported as a special value in the array of synced projects
- Output files written to disk have the following transformations applied:
    - lowercased
    - whitespace replaced with `-`s
    - non-word characters (with the exception of `-` and `_`) removed
- Files that have no text content will no longer be written to disk or included in the js driver file
- The js driver file will only be generated for JSON formats
- Running the CLI with no commands will pull copy from Ditto according to the configuration relative to the current working directory
#### New features
 - Supports `sources.components.folders` property as an array of `{ id: string; name: string }` to restrict component output to components in the specified list of component folders.

### Example
```yaml
sources:
  components: 
    folders:
      - id: folder-1
        name: Folder 1
      - id: folder-2
        name: Folder 2
  projects:
    - name: '[Demo] New User Flow'
      id: 621e71bbb7a68700da355a10
variants: true
format: flat
```

## Context
https://linear.app/dittowords/issue/DIT-3241/update-config

## Test Plan
Testing checklis for default config changes: https://www.notion.so/dittov3/Testing-Checklist-CLI-Defaults-5c1630d7640b4b9b87d34bcaa00db10c

Testing checklist for component folder support:
- [x] Component library is still synced when `sources.components` is set to `true`
- [x] Component library is still synced when `sources.components` is set to `{ enabled: true }`
- [x] Syncing without `format` defined pulls copy in a `flat` format (carryover from last PR)
- [x] Setting `sources.components.folders` to a list of `{id: string; name: string}` and pulling causes only components in those folders to be pulled
- [x] Independent of your configuration, files are always written with the same output structure:
  - [x] `variants`: `true` or `false`
  - [x] `format`: `flat`, `structured`, `android`, `ios-strings`
  - [x] `sources.components`: `true` or undefined
  - [x] `sources.projects`: list of `{id:string; name:string}` or undefined


lib/pull.ts
```diff
@@ -228,115 +229,6 @@ function cleanOutputFiles() {
...
-  let dataString = `module.exports = ${JSON.stringify(data, null, 2)}`
-    // remove quotes around require statements
-    .replace(/"require\((.*)\)"/g, "require($1)");
-
-  const filePath = path.resolve(consts.TEXT_DIR, "index.js");
-  fs.writeFileSync(filePath, dataString, { encoding: "utf8" });
-
-  return `Generated .js SDK driver at ${output.info(filePath)}`;
-}
-
```
azjgard, line 339: Moved to a standalone file.

lib/utils/generateJsDriver.ts
```diff
@@ -0,0 +1,115 @@
+import fs from "fs";
```
azjgard, line 1: Extracted from `lib/pull.ts`

asnewman:
I'm not getting the deprecation notice for these two items 🤔 
image

asnewman (changes requested):
`project add` command seems to be pulling in other projects beyond the one that was selected 
image


`project remove` is also having some strange duplication behavior 🤔 

edit: I think I had the old top level `project` array which cause it to behave like this?

asnewman:
Not sure if this is intended but `project add` with an empty config file allows me to add the component library but then forces me to select a project afterwards (I expected it to be okay with just the component library)

https://user-images.githubusercontent.com/15795866/217672794-c0bab182-438c-4841-b7af-cdb7505b155c.mov

lib/config.ts
```diff
@@ -3,16 +3,30 @@ import path from "path";
...
+  sources: {
+    components: true,
+  },
+  variants: true,
+  format: "flat",
+};
+
+export const DEFAULT_CONFIG = yaml.dump(DEFAULT_CONFIG_JSON);
+
+function createFileIfMissing(filename: string, defaultContents?: any) {
```
asnewman, line 20: Curiosity question - I wonder if it's possible/worth it to do input validation on the config file via zod

lib/init/init.ts
```diff
@@ -29,8 +30,28 @@ export const init = async () => {
...
+        "`projects` is deprecated as a top-level field. Please nest it under `sources` instead."
+      )
+    );
+  }
+  if (hasTopLevelComponentsField) {
+    console.info(
+      output.warnText(
+        "`components` is deprecated as a top-level field. Please nest it under `sources` instead."
+      )
+    );
```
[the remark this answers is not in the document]
asnewman: throwing an error would be my preference!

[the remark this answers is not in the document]
azjgard: I think that's better too! Will update accordingly 👍

lib/config.ts
azjgard: Something I'd love to do in the near future!

asnewman (approved):

PR #47, opened 2023-02-08
[DIT-3310] Revamp file output & add component folder support

azjgard:
## Overview
- Support specifying a list of component folders that should be filtered by
- Refactor file output structure to always be `{source_name}__{variant_name}.{extension}`. This is a major win for simplification, and is only possible as a breaking change and without supporting the `full` format.

## Context
https://linear.app/dittowords/issue/DIT-3310/cli

## Test Plan
- [x] Component library is still synced when `sources.components` is set to `true`
- [x] Component library is still synced when `sources.components` is set to `{ enabled: true }`
- [x] Syncing without `format` defined pulls copy in a `flat` format (carryover from last PR)
- [ ] Setting `sources.components.folders` to a list of `{id: string; name: string}` and pulling causes only components in those folders to be pulled
- [ ] Independent of your configuration, files are always written with the same output structure:
  - [ ] `variants`: `true` or `false`
  - [ ] `format`: `flat`, `structured`, `android`, `ios-strings`
  - [ ] `sources.components`: `true` or undefined
  - [ ] `sources.projects`: list of `{id:string; name:string}` or undefined

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 testing/tmp
 ditto
 bin/
+.env
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -9,7 +9,7 @@ import { Project, ConfigYAML } from "./types";
 
 export const DEFAULT_CONFIG_JSON: ConfigYAML = {
   sources: {
-    components: true,
+    components: { enabled: true },
   },
   variants: true,
   format: "flat",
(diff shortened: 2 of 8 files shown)


asnewman (changes requested):
image

Hmm do you know why it's only pulling variant information?

azjgard:
@asnewman This PR is based on the other one -- since they both have some feedback, for the sake of simplicity in both development and review, I'm just going to merge them together and have a unified description + testing checklist.

PR #51, opened 2023-04-04
Add replace command

asnewman:
## Overview
This PR adds a new `replace` command to search for text and replaces it with Ditto React code.
Note: The `replace` command is not officially supported by Ditto yet.

Usage: `ditto-cli replace <file path> <search string> <replace with>`

## Test Plan

- [ ] run `ditto-cli replace <file path> <search string> <replace with>` on a dummy jsx file and verify it replaces the text with valid Ditto React code.

### DIFF
--- a/lib/ditto.ts
+++ b/lib/ditto.ts
@@ -9,10 +9,16 @@ import { pull } from "./pull";
 import { quit } from "./utils/quit";
 import addProject from "./add-project";
 import removeProject from "./remove-project";
+import { replace } from "./replace";
 
 import processMetaOption from "./utils/processMetaOption";
 
-type Command = "pull" | "project" | "project add" | "project remove";
+type Command =
+  | "pull"
+  | "project"
+  | "project add"
+  | "project remove"
+  | "replace";
 
 const COMMANDS = [
   {
@@ -33,23 +39,29 @@ const COMMANDS = [
       },
     ],
   },
+  {
+    name: "replace",
+    description: "Find and replace Ditto text with code",
+  },
 ] as const;
 
 const setupCommands = () => {
   program.name("ditto-cli");
 
-  COMMANDS.forEach((config) => {
+  COMMANDS.forEach((command) => {
     const cmd = program
-      .command(config.name)
-      .description(config.description)
-      .action(() => executeCommand(config.name));
+      .command(command.name)
+      .description(command.description)
+      .action((str, options) => executeCommand(command.name, options));
 
-    if ("commands" in config) {
-      config.commands.forEach((nestedCommand) => {
+    if ("commands" in command) {
+      command.commands.forEach((nestedCommand) => {
         cmd
           .command(nestedCommand.name)
           .description(nestedCommand.description)
-          .action(() => executeCommand(`${config.name} ${nestedCommand.name}`));
+          .action((str, options) =>
+            executeCommand(`${command.name} ${nestedCommand.name}`, options)
+          );
       });
     }
   });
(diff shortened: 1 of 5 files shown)


lib/ditto.ts
```diff
@@ -62,7 +74,10 @@ const setupOptions = () => {
   );
 };
 
-const executeCommand = async (command: Command | "none"): Promise<void> => {
+const executeCommand = async (
+  command: Command | "none",
+  options: string[]
```
asnewman, line 79: This is the first command with options?

lib/ditto.ts
```diff
@@ -33,23 +39,29 @@ const COMMANDS = [
...
+    name: "replace",
+    description: "Find and replace Ditto text with code",
+  },
 ] as const;
 
 const setupCommands = () => {
   program.name("ditto-cli");
 
-  COMMANDS.forEach((config) => {
+  COMMANDS.forEach((command) => {
```
azjgard, line 51: I still think this variable fits better as being thought of as a "command configuration" or "command group" since we configure nested commands inside of it categorically, but don't feel super strongly!

lib/replace.ts
```diff
@@ -0,0 +1,104 @@
...
+  replaceJSXTextInFile(filePath, { searchString, replaceWith });
+}
+
+function parseOptions(options: string[]): {
+  filePath: string;
+  searchString: string;
+  replaceWith: string;
+} {
+  if (options.length !== 3) {
+    throw new Error("The options array must contain exactly three strings.");
```
azjgard, line 89: Suggestion: could be helpful to ourselves to specify what the three strings should be here

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dittowords/cli",
-  "version": "3.0.0",
+  "version": "3.1.0",
```
azjgard, line 3: Since this doesn't contain any code that is beneficial to our users right now, I'm not sure we should bump the version like this; would be annoying to e.g. get a dependabot PR opened on their repo each time we add a new feature that doesn't actually help them.

I think this should probably be `3.1.0-alpha.0` instead, and we can change to `3.1.0` once all the changes are in and we're ready to release a new public-facing version to support the new plugin. Lmk what you think!

lib/replace.ts
```diff
@@ -0,0 +1,104 @@
...
+      if (regex.test(path.node.value)) {
+        const splitValues = splitByCaseInsensitive(
+          path.node.value,
+          searchString
+        );
+        const nodes: (t.JSXElement | t.JSXText)[] = [];
+
+        splitValues.forEach((splitValue) => {
+          if (splitValue.toLowerCase() === searchString.toLowerCase()) {
+            const identifier = t.jsxIdentifier("DittoComponent");
```
azjgard, line 31: Is it intentional that this only supports `ditto-react` components for now?

azjgard (changes requested):
Looking good! Left a few comments!

lib/ditto.ts
asnewman: Makes sense!

lib/replace.ts
asnewman: Good 👍

package.json
asnewman: Makes sense 👍

lib/replace.ts
asnewman: This is intentional! I'll update to add support for i18next in a separate PR once I get things working end-to-end

azjgard (approved):
Works great!

PR #52, opened 2023-04-05
Add generate-suggestions command

asnewman:
## Overview
This PR adds a new `generate-suggestions` command which will scan through the current repository for strings that match Ditto components. The files must be either `jsx` or `tsx`.
Note: The `generate-suggestions` command is not officially supported by Ditto yet.

Usage: ditto-cli generate-suggestions

Sample output:
image

## Test Plan

- [ ] run ditto-cli generate-suggestions with a dummy jsx file and verify it finds any text that has a 100% match with existing Ditto components.

### DIFF
--- a/lib/generate-suggestions.ts
+++ b/lib/generate-suggestions.ts
@@ -39,48 +39,55 @@ async function findTextInJSXFiles(
     ignore: "**/node_modules/**",
   });
 
+  const promises: Promise<any>[] = [];
+
   for (const file of files) {
-    const code = await fs.readFile(file, "utf-8");
-    const ast = parse(code, {
-      sourceType: "module",
-      plugins: ["jsx", "typescript"],
-    });
+    promises.push(
+      fs.readFile(file, "utf-8").then((code) => {
+        const ast = parse(code, {
+          sourceType: "module",
+          plugins: ["jsx", "typescript"],
+        });
 
-    const occurrences: Occurence[] = [];
+        const occurrences: Occurence[] = [];
 
-    traverse(ast, {
-      JSXText(path) {
-        if (path.node.value.includes(searchString)) {
-          const regex = new RegExp(searchString, "g");
-          let match;
-          while ((match = regex.exec(path.node.value)) !== null) {
-            const lines = path.node.value.slice(0, match.index).split("\n");
+        traverse(ast, {
+          JSXText(path) {
+            if (path.node.value.includes(searchString)) {
+              const regex = new RegExp(searchString, "g");
+              let match;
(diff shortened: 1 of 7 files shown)


lib/generate-suggestions.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type FindResults = { file: string; occurrences: Occurence[] }[];
+
+async function findTextInJSXFiles(
+  path: string,
+  searchString: string
+): Promise<FindResults> {
+  const result: { file: string; occurrences: Occurence[] }[] = [];
+  const files = glob.sync(`${path}/**/*.+(jsx|tsx)`);
```
azjgard, line 36: I think we should probably implicitly ignore `node_modules` since every JavaScript codebase is going to have it

lib/generate-suggestions.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  for (const file of files) {
+    const code = await fs.readFile(file, "utf-8");
```
azjgard, line 39: nit: since order doesn't matter, we could technically do this in a `Promise.all` rather than iteratively in order using a for loop -- not sure if there would be any sort of meaningful performance improvements

lib/generate-suggestions.ts
```diff
@@ -0,0 +1,96 @@
...
+  } = {};
+
+  for (const [compApiId, component] of Object.entries(components)) {
+    if (!results[compApiId]) {
+      results[compApiId] = [];
+    }
+    const result = await findTextInJSXFiles(".", component.text);
+    results[compApiId] = [...results[compApiId], ...result];
+  }
+  console.log(JSON.stringify(results, null, 2));
```
azjgard, line 21: nit: maybe a comment or simple function called `outputToConsole` that just wraps this line to ensure that it's clear as a reader that the `console.log` IS the behavior here

azjgard (changes requested):
Super excited! a couple of small comments

lib/generate-suggestions.ts
asnewman: image

lib/generate-suggestions.ts
asnewman: image

lib/generate-suggestions.ts
asnewman: image

azjgard:
I think it could be good to have this command accept a directory as an argument! Not sure if that would actually be needed by the plugin or not

azjgard (approved):
🌹

PR #53, opened 2023-04-13
Add version option to CLI

asnewman:
This pull request adds a new feature to the ditto-cli command-line tool, allowing users to view the current version of the application by using the -v or --version flag. The version number is retrieved from the package.json file, ensuring that it stays up-to-date with the actual package version.

Changes:

- Import package.json in the main script to access the version number.
- Define a VERSION constant that retrieves the version number from the imported package.json.
- Update the setupOptions() function to use the version() method of the Commander program, which overrides the built-in version option.

These changes make it easier for users to quickly check the installed version of the ditto-cli tool, facilitating better debugging and ensuring they are using the correct version.

### DIFF
--- a/lib/ditto.ts
+++ b/lib/ditto.ts
@@ -4,6 +4,8 @@ import { program } from "commander";
 // to use V8's code cache to speed up instantiation time
 import "v8-compile-cache";
 
+import packageJson from "../package.json";
+
 import { init, needsTokenOrSource } from "./init/init";
 import { pull } from "./pull";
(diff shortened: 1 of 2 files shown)


azjgard (changes requested):
I couldn't get this to work, and get `unknown option '-v'` when trying it. Following up offline!

azjgard (approved):
Confirmed that it works now after publishing and installing an alpha version of latest!

PR #59, opened 2023-05-22
[DIT-4111] Add basic support for iOS .stringsdict

azjgard:
## Overview
Adds explicit support for the iOS format `.stringsdict`.

## Context
https://linear.app/dittowords/issue/DIT-4111/basic-support

This probably isn't particularly helpful for most setups since by default iOS has very particular requirements for their i18n files ("Localizable", in root dir, etc), but we needed this for parity.

## Test Plan
- [x] Ensure you have one or more components with plurals enabled in Ditto
- [ ] Set `sources.components.enabled` to `true` in your `config.yaml`
- [ ] Set `format` as `ios-stringsdict` in `config.yaml`
- [ ] `pull`, and confirm that well-formed `.stringsdict` files are generated in the output directory


README.md
```diff
@@ -227,13 +228,14 @@ If you run the CLI in a directory that does not contain a `ditto/` folder, the f
...
-| **Format**  | **File Extension** |  **Common Usage**   |
-| :---------: | :----------------: | :-----------------: |
-|    flat     |        JSON        |      Web apps       |
-| structured  |        JSON        |      Web apps       |
-|     icu     |        JSON        |      Web apps       |
-|   android   |        XML         | Native Android apps |
-| ios-strings |      STRINGS       |   Native iOS apps   |
+|   **Format**    | **File Extension** |  **Common Usage**   |
+| :-------------: | :----------------: | :-----------------: |
+|      flat       |        JSON        |      Web apps       |
```
asnewman, line 233: What do you think about making these the actual file extensions? (e.g. `.json`, `.xml`, `.strings`, `.stringsdict`). At first glance I would be unsure what STRINGSDICT would generate.

asnewman (approved):

PR #60, opened 2023-05-30
Speed improvement for generate-suggestions

asnewman:
## Overview
Improved the speed of generating suggestions by no longer opening each file for each component. Now we iterate the components for each file.

## Screenshots
Running on the ditto codebase

Previous:
Screen Shot 2023-05-30 at 2 07 47 PM

New:
Screen Shot 2023-05-30 at 2 07 07 PM

## Test Plan

- [ ] Verify `generate-suggestions` still works correctly

### DIFF
--- a/lib/generate-suggestions.ts
+++ b/lib/generate-suggestions.ts
@@ -23,8 +23,6 @@ interface Occurrence {
 
 async function generateSuggestions(flags: { directory?: string }) {
   const components = await fetchComponents();
-
-  console.log("debug3");
   const directory = flags.directory || ".";
 
   const results: { [apiId: string]: Result } = await findComponentsInJSXFiles(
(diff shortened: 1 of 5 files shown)


lib/generate-suggestions.ts
```diff
@@ -22,88 +23,91 @@ interface Occurrence {
...
-              let match;
-              while ((match = regex.exec(path.node.value)) !== null) {
-                const lines = path.node.value.slice(0, match.index).split("\n");
-
-                if (!path.node.loc) {
-                  continue;
-                }
-
-                const lineNumber = path.node.loc.start.line + lines.length - 1;
+            for (const [compApiId, component] of Object.entries(components)) {
```
azjgard, line 58: As discussed on call, we can load components into a map and then get them in `O(1)` time inside of the tree traversal which should speed things up even more!
```jsx
const matchingComponents = componentMap.get(path.node.value);
if (!matchingComponents) return;

azjgard (approved):
My suggestion isn't actually possible, and everything LGTM!

PR #62, opened 2023-06-05
[DIT-4215] generate-suggestions -cf flag

asnewman:
## Overview
Adds the ability to only match components in a given component folder

## Test Plan
- [ ] Run `yarn prepublish`
- [ ] Create a component folder with a component that has text used in the test files
- [ ] Run `generate-suggestions` and verify you get all suggestions
- [ ] Run `generate-suggestions -cf <folder api id>` and verify you get only suggestions for the component in the folder


azjgard (approved):
Works beautifully!
image
image

PR #65, opened 2023-06-13
[DIT-4251] Enable component imports

asnewman:
## Overview
Adds the `import-components` command to import workspace components via a file.

## Test Plan

- [x] Can import ,json files
- [x] Can import .xml
- [x] Can import ios files
- [x] Can import CSV with header information


lib/ditto.ts
```diff
@@ -94,6 +96,31 @@ const COMMANDS: CommandConfig<Command>[] = [
       },
     },
   },
+  {
+    name: "import-components",
+    description:
+      "Import components via a file. For more information please see: https://www.dittowords.com/docs/importing-string-files.",
+    flags: {
+      "-t, --text [value]": {
+        description: "Text column index",
```
azjgard, line 105: I wonder if it makes sense to include something like this for all of the flags here?
```suggestion
        description: "Text column index (.csv format only)",
```

lib/http/importComponents.ts
```diff
@@ -0,0 +1,59 @@
...
+    method: "POST",
+    url: "/v1/components/file",
+    params: {
+      ...(options.csvColumnMapping?.name
+        ? { name: `[${options.csvColumnMapping.name}]` }
+        : {}),
+      ...(options.csvColumnMapping?.text
+        ? { text: options.csvColumnMapping.text }
+        : {}),
+    },
```
azjgard, line 40: Should the other column mappings get included in the request here?

lib/http/importComponents.ts
```diff
@@ -0,0 +1,59 @@
...
+    data: form,
+  };
+
+  try {
+    const { data } = await api(requestOptions);
+
+    return data;
+  } catch (e: any) {
+    console.error("Failed to import file.");
+    console.error(e.response?.data?.message || e.message);
```
azjgard, line 53: Maybe we should parse `error.response.body.code` (assuming that this catch can be triggered by `ApiError`s being thrown on the backend) and have user-friendly copy that is shown rather than the raw error?

asnewman: `e.response?.data?.message` should already be user-friendly copy (at least that's what I've been doing for API errors). `e.message` is a safety net that will just print the code if no user-friendly message is sent from the API. I'm thinking we should make updates to the backend if something is showing up as non-user-friendly, but do you think we need custom messages for the CLI?

azjgard: Ahhhhh you're totally right I forgot about that!

lib/ditto.ts
```diff
@@ -191,6 +218,22 @@ const executeCommand = async (
         ...(options?.lineNumbers ? { lineNumbers: options.lineNumbers } : {}),
       });
     }
+    case "import-components": {
+      if (options.args.length === 0) {
+        console.error("Please provide a file path.");
```
azjgard, line 223: I think it makes sense to write this to stderr, but `console.error` is printing a stack trace as well -- maybe `console.info` or `console.warn` would be better?
image

azjgard (approved):
🔥

lib/ditto.ts
asnewman: Nice suggestion!

PR #66, opened 2023-06-26
[DIT-4270] Support pulling multiple formats at once

azjgard:
## Overview
Adds support for optionally specifying a list of formats in the `format` property.

image
image

## Context
https://linear.app/dittowords/issue/DIT-4270/pull-down-strings-and-stringsdict

## Test Plan
- [x] In `config.yml`, confirm you can pull a single format like `format: flat`
- [x] In `config.yml`, confirm you can pull multiple formats like
```yml
format:
  - ios-strings
  - ios-stringsdict
```

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -21,7 +21,6 @@ export type SupportedFormat =
   | "icu";
 
 export interface ConfigYAML {
-  mode?: "ios";
   sources?: {
     components?: {
       enabled?: boolean;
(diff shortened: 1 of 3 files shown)


lib/types.ts
```diff
@@ -21,6 +21,7 @@ export type SupportedFormat =
   | "icu";
 
 export interface ConfigYAML {
+  mode?: "ios";
```
asnewman, line 24: What is this for?

asnewman (approved):

lib/types.ts
azjgard: oops, left behind from original implementation

PR #70, opened 2023-08-15
[DIT-4720] (v4) Component support overhaul

azjgard:
## Overview
- Write files for component data on a per folder, per variant basis (previously was only on a per variant basis)
```
# Before
ditto_component_library__base.json
ditto_component_library__spanish.json

# After
components__root__base.json
components__root__spanish.json
components__some-folder__base.json
components__some-folder__spanish.json
```
- Gets rid of `sources.components.enabled`, instead supporting `sources.components: true` (for explicitly enabled with no configuration) or implicitly understanding that components should be fetched when `sources.components` is a configuration object
```yml
# Before 1
sources:
  components:
    enabled: true

# Before 2
sources:
  components:
    enabled: true
    folders:
      ...

# After 1
sources:
  components: true

# After 2 (same as Before 2, w/o `enabled`)
sources:
  components:
    folders:
      ...
```
- Adds support for a `sources.components.root` property -- similar to `sources.components`, can be a boolean or a configuration object; this allows the user to explicitly control whether or not components at the root of the component library (i.e. outside of folders) are fetched
```yml
sources:
  components:
    root: true
    folders:
      ...
```

## Context
https://linear.app/dittowords/issue/DIT-4720/cli-allow-users-to-pull-component-folders-into-different-string-files

Should be merged into v4 integration branch.

### Setup
1. Create component folders in your workspace, and populate them with components
2. Populate the root of your component library with components
3. Test against a local API server running on this branch: https://github.com/dittowords/ditto-app/pull/3267
4. `export DITTO_API_HOST=http://localhost:3001` before running `yarn start:api`
### Basic Testing
- [x] Running the CLI with `components.enabled: true` and `components.folders` unspecified writes a component file per variant, with each file following the naming convention `components__root__{variant}.{extension}`
```yml
components:
  enabled: true
```
- [x] Running the CLI with `components.enabled: true`, `components.folders` specified as `{ id: string; name: string }[]`, and `variants: true` writes a component file per variant per component folder, and with each file following the naming convention `components__{folder-name}__{variant}.{extension}`
```yml
components:
  enabled: true
  folders:
    - id: folder-api-id-1
       name: folder-1
    - id: folder-api-id-2
       name: folder-2
variants: true
```
- [x] Running the CLI with `components.enabled: true`, `components.folders` specified as `{ id: string; name: string }[]`, and `variants: false` writes a component file per variant per component folder, and with each file following the naming convention `components__root__{variant}.{extension}`
```yml
components:
  enabled: true
  folders:
    - id: folder-api-id-1
       name: folder-1
    - id: folder-api-id-2
       name: folder-2
variants: false
```
### Root specification
- [x] With neither `sources.components.root` nor `sources.components.folders` specified, confirm that `yarn start` pulls ALL components from your component library
```yml
sources:
  components: true
```
- [x] With `sources.components.root: true` specified and `sources.components.folders` unspecified, confirm that `yarn start` pulls only components from the root of your component library
```yml
sources:
  components:
    root: true
```
- [x] With `sources.components.root` unspecified and `sources.components.folders` specified, confirm that `yarn start` pulls only components from the specified component folders (none from root)
```yml
sources:
  components:
    folders:
      - id: folder-api-id-1
         name: Folder 1
```
- [x] With `sources.components.root: true` and `sources.components.folders` specified, confirm that `yarn start` pulls components from the root of the component library AND components from the specified component folders
```yml
sources:
  components:
    root: true
    folders:
      - id: folder-api-id-1
         name: Folder 1
```
- [x] Set `sources.components.root` to a config object specifying status, and confirm that you only get components from the root that have the status indicated
```yml
sources:
  components:
    root:
      status: WIP
```

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -3,13 +3,12 @@ import path from "path";
 import url from "url";
 import yaml from "js-yaml";
 
-import output from "./output";
 import consts from "./consts";
 import { Project, ConfigYAML, SourceInformation } from "./types";
 
 export const DEFAULT_CONFIG_JSON: ConfigYAML = {
   sources: {
-    components: { enabled: true },
+    components: true,
   },
   variants: true,
   format: "flat",
@@ -220,8 +219,20 @@ function parseSourceInformation(file?: string): SourceInformation {
     validProjects.push(project);
   });
 
-  const shouldFetchComponentLibrary = Boolean(sources?.components?.enabled);
+  const shouldFetchComponentLibrary = Boolean(sources?.components);
+  const componentRoot =
+    typeof sources?.components === "object"
+      ? sources.components.root
+      : undefined;
+  const componentFolders =
+    typeof sources?.components === "object"
+      ? sources.components.folders
+      : undefined;
 
+  /**
+   * If it's not specified to fetch projects or the component library, then there
+   * is no source data to pull.
+   */
   const hasSourceData = !!validProjects.length || shouldFetchComponentLibrary;
 
   return {
@@ -235,8 +246,8 @@ function parseSourceInformation(file?: string): SourceInformation {
     hasTopLevelProjectsField: !!projectsRoot,
     hasTopLevelComponentsField: !!componentsRoot,
     hasComponentLibraryInProjects,
-    componentRoot: sources?.components?.root || false,
-    componentFolders: sources?.components?.folders || null,
+    componentRoot,
+    componentFolders,
   };
 }
 
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -5,7 +5,7 @@ import yaml from "js-yaml";
 
 import output from "./output";
 import consts from "./consts";
-import { Project, ConfigYAML } from "./types";
+import { Project, ConfigYAML, SourceInformation } from "./types";
 
 export const DEFAULT_CONFIG_JSON: ConfigYAML = {
   sources: {
@@ -186,7 +186,7 @@ function dedupeProjectName(projectNames: Set<string>, projectName: string) {
  * - an array of valid, deduped projects
  * - the `variants` and `format` config options
  */
-function parseSourceInformation(file?: string) {
+function parseSourceInformation(file?: string): SourceInformation {
   const {
     sources,
     variants,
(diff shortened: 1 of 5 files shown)


lib/config.ts
```diff
@@ -186,7 +185,7 @@ function dedupeProjectName(projectNames: Set<string>, projectName: string) {
  * - an array of valid, deduped projects
  * - the `variants` and `format` config options
  */
-function parseSourceInformation(file?: string) {
+function parseSourceInformation(file?: string): SourceInformation {
```
jaerod95, line 188: Aren't functions usually best left untyped? Why do we need the explicit type here?

lib/pull.ts
```diff
@@ -358,6 +461,8 @@ async function downloadAndSave(
       });
     }
 
+    // TODO: update this so that all of the separate component library files get spread under one
+    // key, maintaining backwards compatibility for downstream SDKs
```
jaerod95, line 465: Is this TODO meant to be here

lib/utils/generateJsDriver.ts
```diff
@@ -23,6 +23,8 @@ const stringifySourceId = (projectId: string) =>
  */
 
 // TODO: support ESM
+// TODO: to avoid breaking changes with Ditto React, we need to update components part of the file
+// to ...require('folder_file) for each folder into one larger object
```
jaerod95, line 27: Are these TODOs needed?

jaerod95 (commented):
Left a few small questions

lib/config.ts
azjgard: In this case, the explicit type is helpful: when we make changes to `SourceInformation`, we want the type errors to show up here instead of showing up downstream where the implicit return value of this function is used. 

When they show up downstream, we end up having to trace them to this function anyway, so the explicit return definition is helpful for developer experience.

lib/pull.ts
azjgard: Yes, will be implemented in follow-up PR in the integration branch

lib/utils/generateJsDriver.ts
azjgard: Yes, will be implemented in follow-up PR in the integration branch

PR #74, opened 2023-08-17
[DIT-4784] (v4) Generate backwards-compatible index.js driver file for `ditto-react` compat

azjgard:
## Overview
Generate backwards-compatible index.js driver format that includes all component keys for a given variant (independent of the folders they're in) under a single key.

This is done by generating a file that looks like this:
```js
module.exports = {
  ditto_component_library: {
    base: {
      ...require('./components__root__base.json'),
      ...require('./components__folder-1__base.json'),
    },
    some_variant: {
      ...require('./components__root__some_variant.json'),
      ...require('./components__folder-1__some_variant.json'),
    },
  }
}
```

This makes it so that the CLI v4 is backwards compatible with previous `ditto-react` versions.

## Context
https://linear.app/dittowords/issue/DIT-4784/update-indexjs-driver-file-to-be-backwards-compatible-for-ditto-react

Now that components are written on a per-folder basis, we need to update the driver file so that all of the components are still flattened into variant-level properties. This ensures that we maintain backwards compatibility with `ditto-react`.

## Test Plan
1. `export DITTO_API_HOST=http://localhost:3001`
2. Create components in multiple folders with multiple variants in your component library
3. Set up your configuration to pull components
```yml
sources:
  components: true
```
4. Run `yarn start` to pull
5. Confirm that in the generated `index.js` driver file is valid JavaScript, and that you can access all of the components for a given variant via `ditto_component_library[variant-api-id]`

### DIFF
--- a/lib/pull.ts
+++ b/lib/pull.ts
@@ -17,6 +17,8 @@ import {
   Project,
   SupportedFormat,
   ComponentFolder,
+  ComponentSource,
+  Source,
 } from "./types";
 import { fetchVariants } from "./http/fetchVariants";
 import { fetchComponentFolders } from "./http/fetchComponentFolders";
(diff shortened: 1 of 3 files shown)


lib/pull.ts
```diff
@@ -301,6 +303,60 @@ async function downloadAndSave(
 
     const meta = options ? options.meta : {};
 
+    const rootRequest = {
```
azjgard, line 306: All of the code from here through line 354 was pre-existing, and was just moved out of a nested function declaration

rtbarnes (approved):

PR #81, opened 2023-10-12
[DIT-4272] Generate Swift driver file for iOS formats

azjgard:
## Overview
Generates a Swift file to wrap all of the strings generated by the Ditto CLI

## Context
https://linear.app/dittowords/issue/DIT-4272/generate-swift-files-with-static-constants

## Testing
Should be tested via the checklist that lives at https://github.com/dittowords/ditto-app/pull/3565

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -195,6 +195,7 @@ function parseSourceInformation(file?: string): SourceInformation {
     format,
     status,
     richText,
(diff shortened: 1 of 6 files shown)


jaerod95 (approved):

azjgard:
To be included in 4.1 release

PR #86, opened 2023-10-27
[DIT-5321] Support generating ESM JavaScript driver

azjgard:
## Overview
Refactors `generateJsDriver` to depend on the `type` property of the nearest `package.json` file.
- if `type` is `module`, an ESM module is generated
- otherwise, a CommonJS module is generated 

## Context
https://linear.app/dittowords/issue/DIT-5321/make-ditto-react-work-with-default-vite-react-setup

## Test Plan
- [ ] Test with https://github.com/dittowords/ditto-react-demo/pull/10 and confirm that the sample project operates normally

### DIFF
--- a/lib/utils/determineModuleType.ts
+++ b/lib/utils/determineModuleType.ts
@@ -51,5 +51,5 @@ function getTypeOrDefault(value: string | null): "commonjs" | "module" {
     return valueLower;
   }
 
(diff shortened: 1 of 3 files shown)


asnewman (approved):

azjgard:
To be included in 4.1 release

PR #88, opened 2023-11-13
Add sample-data field to the components-folder command

asnewman:
## Overview
Adds a flag to return sample data for the components-folder command.

## Test Plan

- [ ] `yarn start components-folder` returns components folder without sample-components
- [ ] `yarn start components-folder --sample-data` returns components folder with sample-components

### DIFF
--- a/lib/component-folders.ts
+++ b/lib/component-folders.ts
@@ -1,7 +1,7 @@
 import { fetchComponentFolders } from "./http/fetchComponentFolders";
 
-async function showComponentFolders() {
-  const folders = await fetchComponentFolders();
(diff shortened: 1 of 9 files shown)


azjgard (approved):
Pushed a couple of commits that would be godo for you to review!
- Added typechecking to the package.json commands to help us catch errors at compile time instead of allowing us to run with them
- Passed empty options object in the place where `fetchComponentFolders` is called as part of the `pull` command
- Fixed a couple of miscellaneous type errors
- Fixed a typo in parsing the flag when running the `component-folders` command

PR #89, opened 2023-11-16
Only run init() on commands that need it

asnewman:
## Overview
This PR makes a change to take into account what command is being run before deciding to run init() or not.

Also includes two minor improvements:
- Explicitly use `structured` format when querying for components
- Ignore white space only lines of code and components when generating suggestions

## Context
The Ditto VSCode extension runs commands such as `generate-suggestions`, `component-folders`, and `replace`. These commands don't rely on local data (it calls the Ditto API) and thus doesn't require a `config.yml` to be present. Because of this, it is unnecessary to run init() when these commands are called.

## Test Plan

- [x] Delete `ditto` directory
- [x] Run `yarn start generate-suggestions`, verify that the `ditto` directory isn't created
- [x] Run `yarn start pull`, verify that the `ditto` directory is created
- [x] Update `testfiles/test1.jsx` to include text for a component in your workspace. Verify `yarn start generate-suggestions` results in a suggestion for that text

### DIFF
--- a/lib/ditto.ts
+++ b/lib/ditto.ts
@@ -152,7 +152,7 @@ const setupCommands = () => {
           } else {
             cmd.option(flags, description);
           }
-        }
+        },
       );
     }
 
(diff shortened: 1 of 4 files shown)


lib/generate-suggestions.ts
```diff
@@ -77,11 +79,16 @@ async function findComponentsInJSXFiles(params: {
                 };
               }
 
-              if (path.node.value.includes(component.text)) {
+              if (
+                // Skip white space lines
+                !/^\s*$/.test(path.node.value) &&
+                !/^\s*$/.test(component.text) &&
```
azjgard, line 85: is this change meant to be included in this PR?

lib/ditto.ts
```diff
@@ -177,16 +177,25 @@ const setupCommands = () => {
...
+  options: any,
 ): Promise<void> => {
-  const needsInitialization = needsTokenOrSource();
+  const configFileReliantCommands = [
+    "pull",
+    "none",
+    "project",
+    "project add",
+    "project remove",
+  ];
```
azjgard, line 195: nit: could move these commands to top of file as a constant

azjgard (approved):

PR #100, opened 2024-02-28
[DIT-6177] Add pull --sample-data flag for variants

asnewman:
## Overview
This PR adds the `--sample-data` flag to the `pull` command and is intended to allow support for sample data to be included in the `pull` execution. With this PR, only variants will be supported.

## Test Plan

- [ ] Run `yarn pull --sample-data` and verify sample variant data is included

### DIFF
--- a/lib/pull.ts
+++ b/lib/pull.ts
@@ -583,7 +583,7 @@ export interface PullOptions {
 
 export const pull = async (options?: PullOptions) => {
   const meta = options ? options.meta : {};
-  const includeSampleData = options ? options.includeSampleData : false;
+  const includeSampleData = options?.includeSampleData || false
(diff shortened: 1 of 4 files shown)


lib/pull.ts
```diff
@@ -578,15 +578,20 @@ async function downloadAndSave(
 
 export interface PullOptions {
   meta?: Record<string, string>;
+  includeSampleData?: boolean;
 }
 
 export const pull = async (options?: PullOptions) => {
   const meta = options ? options.meta : {};
+  const includeSampleData = options ? options.includeSampleData : false;
```
azjgard, line 586: nitty
```suggestion
  const includeSampleData = options?.includeSampleData || false
```

azjgard (approved):

PR #105, opened 2024-03-22
[DIT-6379] Fix terminal hanging when Ctrl + C pressed during `pull` and token prompt

azjgard:
## Overview
Fixes an issue where the CLI would become unresponsive but fail to fully exit when the user attempted to use Ctrl + C to bail out of the process.

## Context
https://linear.app/dittowords/issue/DIT-6379/fix-cli-hanging-when-ctrlc-out-of-token-collection-prompt

## Test Plan
1. Remove the file `~/.config/ditto`
2. Run `yarn start pull`
3. At the prompt, hit Ctrl + C
4. Confirm that the CLI exits and control is immediately returned to the user

### DIFF
--- a/lib/utils/quit.ts
+++ b/lib/utils/quit.ts
@@ -1,4 +1,4 @@
 export function quit(message: string | null, exitCode = 2) {
   if (message) console.log(`\n${message}\n`);
-  process.exitCode = exitCode;
+  process.exit(exitCode);
 }
--- a/lib/init/token.ts
+++ b/lib/init/token.ts
@@ -1,4 +1,5 @@
 import fs from "fs";
+import * as Sentry from "@sentry/node";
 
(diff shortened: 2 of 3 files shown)


lib/utils/quit.ts
```diff
@@ -1,4 +1,4 @@
 export function quit(message: string | null, exitCode = 2) {
   if (message) console.log(`\n${message}\n`);
-  process.exitCode = exitCode;
```
azjgard, line 3: This was incorrect - setting the exit code ensured that when the process exited of other means, it would have the specified status, but the intent of the `quit()` function in all instances of usage is to immediately terminate the running process.

jaerod95 (approved):
Tested the exit functionality, everything seems to be working as expected

PR #114, opened 2025-04-07
Cleanup build scripts and upgrade typescript and esbuild

jaerod95:
## Overview

Upgrades typescript and esbuild to latest versions and create a modern build script for bundling the CLI. This removes the need for the `esbuild-node-tsc` package which has been removed from the dev dependnencies list. The `prepublish`, `start`, and `sync` scripts have been updated to reflect the new esbuild file.
Additionally, drops the use of `v8-compile-cache` to simplify dependencies as the benefits for a cli of this size were minimal at best, and adds a basic `CONTRIBUTING.md` file to house future instructions to describe building and running the CLI locally.

## Test Plan

- [ ] install new dependencies with `yarn install`
- [ ] `DITTO_API_HOST=http://localhost:3001 yarn start` builds and runs the plugin application like normal.


CONTRIBUTING.md
```diff
@@ -0,0 +1,19 @@
...
+
+Prerequisites:
+
+- Node v20 or above
+- yarn (1.x)
+
+To compile and run the cli, use the `yarn start` command. Any additional parameters the CLI accepts can be added. For example:
+
+- `yarn start pull`
+- `yarn start --version`
```
marla-hoggard, line 17: This might be a good spot to mention the `DITTO_API_HOST` variable as well.

package.json
```diff
@@ -3,13 +3,12 @@
...
   "license": "MIT",
-  "main": "bin/index.js",
+  "main": "bin/ditto.js",
   "scripts": {
-    "prepublishOnly": "ENV=production etsc && sentry-cli sourcemaps inject ./bin && npx sentry-cli sourcemaps upload ./bin --release=\"$(cat package.json | jq -r '.version')\"",
+    "prepublishOnly": "ENV=production node esbuild.mjs && sentry-cli sourcemaps inject ./bin && npx sentry-cli sourcemaps upload ./bin --release=\"$(cat package.json | jq -r '.version')\"",
     "prepare": "husky install",
-    "start": "tsc --noEmit --excludeFiles './**/*.test.ts' && etsc && node bin/ditto.js",
-    "sync": "tsc --noEmit --excludeFiles './**/*.test.ts' && etsc && node bin/ditto.js pull",
-    "dev": "tsc --noEmit --excludeFiles './**/*.test.ts' && etsc --watch"
```
marla-hoggard, line 12: Just double checking - did you mean to remove `dev`? Do we lose anything important from removing it?

tsconfig.json
```diff
@@ -1,16 +1,21 @@
...
   "compilerOptions": {
-    "allowJs": true,
+    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string"],
+    "module": "commonjs",
+    "target": "es2019",
+
+    "strict": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
```
marla-hoggard, line 10: Sounds like a nice benefit, but is there any risk to adding this? Such as to the files we generate via the cli?

marla-hoggard (approved):
I don't have a huge amount of context for these changes, but I was able to run `yarn start` `yarn sync` and `yarn prepublishOnly` successfully. prepublishOnly exits with an error, but the same is true on the parent branch, so I think that means it's expected?

tsconfig.json
jaerod95: These are the defaults provided by the "extends": "@tsconfig/node16/tsconfig.json". Instead of relying on that package that is poorly maintained, it makes more sense to just set the fields ourselves to remove any ambiguity.

CONTRIBUTING.md
jaerod95: This is an internal variable that isn't applicable in the public context in which this repo is provided.

tsconfig.json
marla-hoggard: Just specifically asking about `forceConsistentCasingInFileNames` - since the CLI generates files, i didn't know if there was any risk in adding this which wasn't in our previous config.

CONTRIBUTING.md
marla-hoggard: Isn't this doc for contributors - people who would submit a PR? Wouldn't they have to know how to run the app locally?

tsconfig.json
jaerod95: `forceConsistentCasingInFileNames` was already being set to true as it is set in the node16 defaults i mentioned above: 
Screenshot 2025-04-08 at 2 54 22 PM

jaerod95: As for concerns with generating files, this tsconfig is only used while developing the cli. Using the CLI knows nothing about the tsconfig as it is an internal configuration.

PR #116, opened 2025-04-10
Jrod/dit 10045 reintroduce pull command with ns cli

jaerod95:
## Overview

Restores support for parsing API tokens via a global config file.

## Test Plan

- [ ] Deleting your ~/.config/ditto file and then running the CLI should walk you through fetching and saving your api token.
- [ ] Entering an invalid API Token should reprompt you to input a valid api token.
- [ ] It should generate json files in your ./ditto/ directory in the repo when running pull, and pull down all projects with the base variant by default. This means a file for each project with the name of `project-id___base.json`
- [ ] Setting some variants into your configuration should generate a different variant file per variant per project if variants exist for text items in that project for provided variants
- [ ] Filtering by project ids should only fetch relevant text from projects.


lib/src/outputs/shared.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from "zod";
+
+/**
+ * These filters that are common to all outputs, used to filter the text items that are fetched from the API.
+ * They are all optional by defualt unless otherwise specified in the output config.
+ */
+export const ZBaseOutputFilters = z.object({
+  projects: z.array(z.string()).optional(),
+  variants: z.array(z.string()).optional(),
+});
```
marla-hoggard, line 10: Same comment I made in another place - why are we setting up the config to only ever take ids, when we set up the endpoints to ultimately support more complex filtering?

lib/src/formatters/json.ts
```diff
@@ -0,0 +1,92 @@
...
+    const variablesOutputFile = new JSONOutputFile(
+      "variables",
+      appContext.projectConfigDir
+    );
+
+    for (let i = 0; i < data.textItems.length; i++) {
+      const textItem = data.textItems[i];
+
+      outputJsonFiles[textItem.projectId] ??= new JSONOutputFile(
+        `${textItem.projectId}___${textItem.variantId || "base"}`,
```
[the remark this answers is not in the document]
jaerod95: we'll need to tackle that still. since it's just a few workspaces it's been lower priority

lib/src/formatters/mixins/javascriptCodegenMixin.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export default function javascriptCodegenMixin<TBase extends Constructor>(
+  Base: TBase
+) {
+  return class JavascriptCodegenHelpers extends Base {
+    protected indentSpaces: number = 2;
+
+    protected sanitizeStringForJSVariableName(str: string) {
+      return str.replace(/[^a-zA-Z0-9]/g, "_");
```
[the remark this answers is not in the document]
jaerod95: yes they have to be in order for it to be a valid javascript variable name.

lib/src/formatters/json.ts
```diff
@@ -0,0 +1,92 @@
...
+
+type JSONAPIData = {
+  textItems: TextItemsResponse;
+  variablesById: Record<string, VariablesResponse[number]>;
+};
+
+export default class JSONFormatter extends applyMixins(
+  BaseFormatter<JSONAPIData>) {
+
+  protected async fetchAPIData() {
```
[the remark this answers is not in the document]
jaerod95: woof possibly. I wrote it this way to make intellisense easier to deal with but we'll need to look into that.

lib/src/formatters/shared/fileTypes/OutputFile.ts
```diff
@@ -0,0 +1,33 @@
...
+  ) {
+    this.filename = filename;
+    this.path = path;
+    this.extension = extension;
+    this.content = content;
+    this.metadata = metadata;
+  }
+
+  get fullPath() {
+    return `${this.path}/${this.filename}.${this.extension}`;
```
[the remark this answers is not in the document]
jaerod95: since we define it and will test it we should be ok here.

lib/src/formatters/mixins/javascriptCodegenMixin.ts
```diff
@@ -0,0 +1,44 @@
...
+          return m.name;
+        })
+        .sort()
+        .join(", ");
+
+      return `import { ${formattedModules} } from "${moduleName}";`;
+    }
+
+    protected codegenDefaultImport(module: string, moduleName: string) {
+      return `import ${module} from "${moduleName}";\n`;
```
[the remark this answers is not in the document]
jaerod95: Added it in

lib/src/formatters/mixins/javascriptCodegenMixin.ts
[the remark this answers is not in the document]
marla-hoggard: hmm, i wonder why we picked that as our default delimiter then...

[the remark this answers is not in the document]
jaerod95: This is something totally different, if we were to try to remove all invalid characters from all programming languages we would have no possible string values.

lib/src/services/apiToken.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    const eventId = Sentry.captureException(error);
+    const eventStr = `\n\nError ID: ${logger.info(eventId)}`;
+
+    return await quit(
+      logger.errorText(
+        "Something went wrong. Please contact support or try again later."
+      ) + eventStr
+    );
```
marla-hoggard, line 71: We do this same sentry code in several places. Might be nice to make a helper for it:

```typescript
export const quitWithError = async (error: unknown, message?: string) => {
  const eventId = Sentry.captureException(error);
  const eventStr = `\n\nError ID: ${logger.info(eventId)}`;
  const errorMessage = message || "Something went wrong. Please try again later.";

  return await quit(logger.errorText(errorMessage) + eventStr);
};
```

Then lines 64-71 could simply be:
```typescript
return await quitWithError(error);
```

lib/src/services/globalConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+      token: z.string(),
+    })
+  )
+);
+
+type GlobalConfigYAML = z.infer<typeof ZGlobalConfigYAML>;
+
+/**
+ * Read data from a global config file
+ * @param file defaults to `CONFIG_FILE` defined in `constants.js`
```
marla-hoggard, line 20: Looks like this is not true anymore.

lib/src/services/globalConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param defaultData defaults to `{}`
```
marla-hoggard, line 21: It took me a minute to understand what we did with `defaultData`. Perhaps rename to `fallbackData` and then update this comment to explain it's what's returned if parsing the yaml file fails.

Or just delete it. There are no calls that provide the second parameter.

lib/src/services/globalConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Save a token to the global config file
+ * @param file The path to the global config file
+ * @param hostname The hostname to save the token for
+ * @param token The token to save
+ */
+export function saveToken(file: string, hostname: string, token: string) {
+  const data = readGlobalConfigData(file);
+  data[hostname] = []; // only allow one token per host
+  data[hostname][0] = { token };
```
marla-hoggard, line 54: Nit: You can do this on one line, which I think is a bit more readable.
```suggestion
  data[hostname] = [{ token }]; // only allow one token per host
```

lib/src/services/projectConfig.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+export async function initProjectConfig() {
+  const projectConfig = readProjectConfigData();
+  appContext.setProjectConfig(projectConfig);
+}
+
+/**
+ * Read data from a global config file
+ * @param file defaults to `CONFIG_FILE` defined in `constants.js`
```
marla-hoggard, line 32: Same here - this is no longer the default

marla-hoggard (changes requested):
Finished the code review! Comments are mostly small/nits, but the APIToken helpers need another pass, as the usage of the `token` parameter is inconsistent and not quite right (and currently includes type errors). I'll wait for you to update that set of methods before testing.

lib/src/services/globalConfig.ts
jaerod95: This is very important to include as it helps things not crash when the user has errors in their config file. If there is a typo in their config file for example, this lets us reprompt for an api key instead of just crashing.

lib/src/http/checkToken.ts
```diff
@@ -0,0 +1,65 @@
+import httpClient, { defaultInterceptor } from "./client";
```
marla-hoggard, line 1:
```suggestion
 { defaultInterceptor } from "./client";
```

lib/src/services/globalConfig.ts
marla-hoggard: There are 8 `readGlobalConfigData(file)` calls in the codebase. None of them provide a value for the second parameter. I'm not suggesting that we don't return something, I'm saying that there are no instances that ever actually provide a value for `defaultData`. So, unless there's a clear use case where we intend to provide custom fallback data, there's no reason to overcomplicate this method by accepting a parameter we never provide. Just make the method return `{}` if the config fails validation.

lib/src/http/client.ts
```diff
@@ -0,0 +1,18 @@
+import axios, { AxiosRequestConfig, InternalAxiosRequestConfig } from "axios";
```
marla-hoggard, line 1:
```suggestion
import axios, { InternalAxiosRequestConfig } from "axios";
```

lib/src/services/apiToken.ts
```diff
@@ -0,0 +1,144 @@
...
+import checkToken from "../http/checkToken";
+import logger from "../utils/logger";
+import { quit } from "../utils/quit";
+import * as Sentry from "@sentry/node";
+import { prompt } from "enquirer";
+
+export async function initAPIToken(
+  token: string | undefined = appContext.apiToken,
+  configFile: string = appContext.configFile,
+  host: string = appContext.apiHost
```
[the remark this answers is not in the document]
jaerod95: This format of params with defaults is meant to aide when it comes to being able to test this method easily. It's a stylistic choice that already existed in the legacy implementation, and since this was a copy then modify scenario from the legacy codebase, it's included. I personally like the concept for testing as long as it's a consistent theme throughout the the implementation. For now, not planning on updating it.

lib/src/services/globalConfig.ts
jaerod95: This was done to provide an easy way to inject values when writing test files in the legacy implementation. I'll just remove it in the newer implementation

marla-hoggard: I'm all for adding some extra things today that we know will be helpful in the near future for functionality we plan to add. But extra parameters that will never be used just add to code complexity and to the test burden.

If it's hard for us to test those appContext values, then I see how providing them to other methods as parameters could make testing easier. But in this case, it's just an extra code path that we don't need.

jaerod95: Again, this was copied from legacy, happy to remove

lib/src/http/checkToken.ts
```diff
@@ -0,0 +1,65 @@
+ { defaultInterceptor } from "./client";
```
marla-hoggard, line 1: Looks like this suggestion got mangled.
```suggestion
import { defaultInterceptor } from "./client";
```

marla-hoggard (approved):
I've done a bunch of testing and things are working well! I do have one piece of feedback, which we could fix here or as a followup. The UX for inputting an invalid api key is a bit strange. When I enter a bad value, my initial line updates to say "token (value)" and then the new prompt is already pre-populated with my bad value and then an error message two lines below where my cursor is. It's just a bit confusing/overwhelming.

This is the state of the terminal immediately after I hit submit on "bad.key" the first time
Screenshot 2025-04-15 at 5 14 43 PM

PR #117, opened 2025-04-16
Jrod/dit 10122 add support for c flag for defining the location of the

jaerod95:
## Overview

Adds support for passing in a `-c, --config <path_to_config>` flag into the pull command for specifying where to find the cli config.yml.
Additionally, cleans up the file structure to better match the internal architecture used within the CLI.
Adds support for new `outDir` field to specify which folder you would like to have the cli output to at the global and per output level.
Finally, supports overriding project/variant configs at the output level.

## Test Plan

- [ ] change the name of your config.yml file to something else. Call the pull command passing the flag `-c ./relative/path/to/your/renamed/config.yml` and ensure it reads the config correctly
- [ ] Specify an outDir property at the project level. it should output the files generated to the relative path provided.
- [ ] Add project filters at the output level and ensure that is respected over the global project configs value. and empty array should still override what is defined globally.


marla-hoggard (changes requested):
Everything is working great except for specifying an `outDir` property at the top level of the yml file. Project/variant overrides are working nicely from top and per-output specification. outDir works great if provided to a specific output. I can provide different directories for each output. and each directory is being made for me if it didn't exist.

Just can't specific an `outDir` at the top level of the config file. When I do that, it's being ignored and saving to the default location.

marla-hoggard (approved):

PR #118, opened 2025-04-16
[DIT-10076] Unit tests for API Token service methods

marla-hoggard:
## Overview

Writes unit tests for each of the api token service methods.

Since these methods call each other, I ran into to issues with proper mocking when they were all in one file, so I reorganized to put each method in its own file, and created a test file for each.

Generally speaking, each method was just copy/pasted into the new file, but I did make a couple updates:
- Removed all the parameters from `initAPIToken` - they were only added for the purpose of testing and it turns out I didn't need them
- I moved the `validate` method we pass into `prompt` to a named function so it could be tested

## Context

https://linear.app/dittowords/issue/DIT-10076/add-tests-for-apitoken-service

## Test Plan

- [ ] The tests should pass in CI
- [ ] The tests should pass locally by running `npx jest apiToken` (to run all the tests in the new directory)
- [ ] How do they look?


jest.config.ts
```diff
@@ -11,6 +11,7 @@ const config: Config = {
   ],
   watchPathIgnorePatterns: ["<rootDir>/.testing/", "<rootDir>/testing/"],
   collectCoverageFrom: ["lib/**/*.{js,jsx,ts,tsx}"],
+  restoreMocks: true,
```
marla-hoggard, line 14: In general, I still have calls in each test file for this, but apparently you can set this here to automatically restore. Seemed like a nice catch-all in case we forget somewhere.

lib/src/services/apiToken/collectAndSaveToken.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  const token = "token";
+  const host = "host";
+  const apiHost = "apiHost";
+  const sanitizedHost = "hostname";
+
+  beforeEach(() => {
+    priorToken = appContext.apiToken;
+    priorHost = appContext.apiHost;
+    appContext.setApiToken("");
+    appContext.apiHost = apiHost;
```
marla-hoggard, line 26: For the functions that use `appContext` values that we want to specify, i'm saving the original value, changing it to what we want, then restoring the prior value after the test. This might be total overkill, but it seemed like a nice safety measure. Thoughts?

lib/src/services/apiToken/collectToken.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    jest.restoreAllMocks();
+  });
+
+  it("prompts for API token and returns it", async () => {
+    const response = await collectToken();
+    expect(response).toBe(token);
+    expect(logger.url).toHaveBeenCalled();
+    expect(logger.bold).toHaveBeenCalled();
+    expect(logger.info).toHaveBeenCalled();
+    expect(logger.writeLine).toHaveBeenCalled();
```
marla-hoggard, line 33: Is this overkill? Should i ditch the logger checks? Does it make the test too brittle/resistent to change?

lib/src/services/apiToken/initAPIToken.ts
```diff
@@ -0,0 +1,33 @@
+import appContext from "../../utils/appContext";
+import fs from "fs";
+import * as configService from "../globalConfig";
+import collectAndSaveToken from "./collectAndSaveToken";
+import validateToken from "./validateToken";
+import getURLHostname from "./getURLHostname";
+
```
marla-hoggard, line 7: This method was updated to remove the optional parameters (never used in the real code) and hard-code to the `appContext` values. Would be a good idea to sanity check me here.

lib/src/services/apiToken/promptForApiToken.ts
```diff
@@ -0,0 +1,26 @@
+import checkToken from "../../http/checkToken";
+import { prompt } from "enquirer";
+
+export const validate = async (token: string) => {
+  const result = await checkToken(token);
+  if (!result.success) {
+    return result.output?.join("\n") || "Invalid API key";
+  }
+  return true;
+};
```
marla-hoggard, line 10: I moved the `validate` file to its own saved function so it could be tested. It's identical to what was previously defined inline on line 22.

lib/src/services/apiToken/collectAndSaveToken.test.ts
jaerod95: I think this is fine for now.

lib/src/services/apiToken/collectToken.test.ts
jaerod95: I think it's okay to be brittle here as this will rarely change, and it does help us know that the messages were logged out. if the messages weren't getting logged out, that would be useful to know.

lib/src/services/apiToken/initAPIToken.ts
jaerod95: it's fine to remove the optional parameters, it should all still work the same.

PR #122, opened 2025-04-22
[DIT-10167] Update version to 5.0.0-beta.6 and add Vue I18n framework support in …

jaerod95:
## Overview

Adds support for the vue-i18n framework. Adds support for reformatting double mustaches with single mustaches to be compatible (e.g. `{{ variable_name }}` to `{ variable_name }`

## Test Plan

- [ ] Add `vue-i18n` to your config.yml. It should generate json files exactly like i18next but with single `{` `}` mustaches around variables.

### DIFF
--- a/lib/src/formatters/frameworks/json/vue-i18n.ts
+++ b/lib/src/formatters/frameworks/json/vue-i18n.ts
@@ -1,0 +1,135 @@
+import JavascriptOutputFile from "../../shared/fileTypes/JavascriptOutputFile";
+import OutputFile from "../../shared/fileTypes/OutputFile";
+import { applyMixins } from "../../shared";
+import javascriptCodegenMixin from "../../mixins/javascriptCodegenMixin";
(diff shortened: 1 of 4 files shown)


lib/src/formatters/frameworks/json/vue-i18n.ts
```diff
@@ -0,0 +1,135 @@
...
+      for (const key in file.content) {
+        const content = file.content[key];
+        if (typeof content === "string") {
+          file.content[key] = content
+            .replaceAll(/{{/g, "{")
+            .replaceAll(/}}/g, "}");
+        }
+      }
+    }
+
```
asnewman, line 26: Am I correct in assuming that this is the only difference between the `i18next` formatter and the `vue-i18n` formatter?

asnewman (approved):

lib/src/formatters/frameworks/json/vue-i18n.ts
jaerod95: yup!

PR #123, opened 2025-08-08
Add rich text support for CLI

asnewman:
## Summary
Adds support for fetching and outputting rich text (HTML) content from the Ditto API.

## Usage
Add `richText: html` to your output configuration:

```yaml
outputs:
  - format: json
    framework: i18next
    richText: html
```

When enabled, the CLI will output HTML-formatted text values in JSON files, falling back to plain text when rich text is unavailable.

Closes DIT-11112

### DIFF
--- a/lib/src/http/textItems.ts
+++ b/lib/src/http/textItems.ts
@@ -11,6 +11,7 @@ const TextItemsResponse = z.array(
   z.object({
     id: z.string(),
     text: z.string(),
+    richText: z.string().optional(),
     status: z.string(),
     notes: z.string(),
     tags: z.array(z.string()),
--- a/lib/src/http/textItems.ts
+++ b/lib/src/http/textItems.ts
@@ -5,6 +5,7 @@ import { z } from "zod";
 export interface PullFilters {
   projects?: { id: string }[];
   variants?: { id: string }[];
+  richText?: "html";
 }
 
 const TextItemsResponse = z.array(
@@ -25,10 +26,19 @@ export type TextItemsResponse = z.infer<typeof TextItemsResponse>;
 
 export default async function fetchText(filters?: PullFilters) {
   try {
+    const params: { filter: string; richText?: string } = {
+      filter: JSON.stringify({
+        projects: filters?.projects,
+        variants: filters?.variants,
+      }),
+    };
+    
+    if (filters?.richText) {
+      params.richText = filters.richText;
+    }
+
     const response = await httpClient.get("/v2/textItems", {
-      params: {
-        filter: JSON.stringify(filters),
-      },
+      params,
     });
 
     return TextItemsResponse.parse(response.data);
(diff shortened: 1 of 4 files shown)


lib/src/http/textItems.ts
```diff
@@ -24,10 +26,15 @@ export type TextItemsResponse = z.infer<typeof TextItemsResponse>;
 
 export default async function fetchText(filters?: PullFilters) {
```
marla-hoggard, line 27: The changes in this file break the existing filter functionality. The project and variant filters are no longer being respected. They still need to be passed in as stringified to the filter param the way the were before, while the `richText` value is a separate top-level field.

Here's an example of the output we ultimately want, which results in successful results from the public API:
`{{baseUrl}}/v2/textItems?filter={"projects": [{"id": "testing" }], "variants": [{"id": "all" }] }&richText=html`.

Additionally, we're directly mutating the `filters` parameter (which is marked as optional but as far as I cant tell should be required) via the `delete` call due to `const params = filters` giving both variables the same reference object. This could cause unintended side effects.

If we want to make the changes just in this file, we could refactor to this:

```typescript
const { richText, ...filter } = filters;

const params =
      richText === false
        ? { filter: JSON.stringify(filter) }
        : { filter: JSON.stringify(filter), richText };
```

However, it may be better to refactor this higher up, so that we aren't receiving `richText` mixed in with the other `filters`. We could instead either refactor `this.generatePullFilter` to return the params split up the way we need them, or we could remove `richText` from `this.generatePullFilter` entirely, and make a separate method for generating additional query params. Either way, we would then pass the filter and richText values in to `fetchText` as already separated.

marla-hoggard (changes requested):
I'm going to push up a few commits to fix the filter issue, as well as update a prompt message that i noticed was out of date.

lib/src/services/apiToken/collectToken.ts
```diff
@@ -6,9 +6,8 @@ import promptForApiToken from "./promptForApiToken";
  * @returns The collected token
  */
 export default async function collectToken() {
-  const apiUrl = logger.url("https://app.dittowords.com/account/devtools");
```
marla-hoggard, line 9: While trying to get set up, I noticed that this URL is out of date, so i updated it.

lib/src/http/textItems.ts
```diff
@@ -7,10 +7,16 @@ export interface PullFilters {
   variants?: { id: string }[];
 }
 
+export interface PullQueryParams {
+  filter: string; // Stringified PullFilters
+  richText?: "html";
+}
```
marla-hoggard, line 13: If we convert these to zod types, we could probably enforce the filter type is actually a stringified object with the right format. But I don't think the TS inferred type would enforce it, so it would only help if we ran zod.parse, which is probably not worth it -- the API endpoint should throw if it's wrong. It would only help us if we were making changes and then tested locally. I'm thinking stronger unit tests are probably a better move than that.

marla-hoggard: Also, this is kind of an odd location for PullFilters and PullQueryParams types to live, but when I originally moved them to the `json` file where they're also used, that creates a circular import, so I left them here for now.

PR #125, opened 2025-09-29
Bump beta version to 7

JWhite30515:
## Overview

Bumps beta version to 7 in `package.json`, we missed doing so for richText release


marla-hoggard (approved):
Approving but don't merge it yet. Gotta make sure we know the right way to do the release tagging.

PR #136, opened 2026-03-27
Upgrade zod from v3 to v4

laurakoye:
## Overview

Update zod 3 to zod 4

  - Replace .merge() with .extend() (deprecated in v4)
  - Replace .strict() with z.strictObject() (deprecated in v4)
  - Add explicit key schema to z.record() (single-arg no longer supported)
  - Make z.undefined() fields optional to preserve v3 behavior
  - Replaced .flatten() with prettifyError() for displaying error messages

## Screenshots
Screenshot 2026-03-27 at 9 35 48 AM

## Test Plan

- [ ] yarn test
- [ ] yarn sync

### DIFF
--- a/lib/src/formatters/shared/base.ts
+++ b/lib/src/formatters/shared/base.ts
@@ -130,7 +130,7 @@ export default class BaseFormatter<OutputFileType, APIDataType = unknown> {
 
   public async format(): Promise<void> {
     const data = await this.fetchAPIData();
-    const files = await this.transformAPIData(data);
+    const files = this.transformAPIData(data);
     await this.writeFiles(files);
   }
 
--- a/lib/src/http/variables.ts
+++ b/lib/src/http/variables.ts
@@ -9,49 +9,39 @@ const ZBaseVariable = z.object({
   name: z.string(),
 });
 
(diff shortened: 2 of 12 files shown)


marla-hoggard (commented):
In addition to the zod-related changes, we'll also need to bump the version number of the CLI itself to set up the new release of the CLI. Since it's just an internal change, we can do a patch update to v5.4.1. Although keep an eye out, as there is another open PR that bumps to v5.5.0. If that goes in first, this will have to take us up to v5.5.1.

In case you need it, here's the full instructions for making CLI updates, including all the steps that will come after this PR: https://www.notion.so/dittov3/Working-on-Ditto-Dev-Tools-ad7b4a133c4b4327961c0af4ba2c7637

marla-hoggard (commented):
As far as I can tell, without deep knowledge of all the different export formats, everything is looking good/consistent with master. See my comments in the code - there's a few more things we can/should update. And then we'll need to bump the package version before this can merge.

bparrish17 (approved):
tested a bunch of outputs and everything looks good to me 👍 thanks for getting this across the finish line for laura

PR #139, opened 2026-06-05
[DIT-13028] Connect CLI to the public API scan endpoints

joustrich:
## Overview

Connect CLI to the public API scan endpoints

## Context

Note that this PR pairs with [this ditto-app PR](https://github.com/dittowords/ditto-app/pull/8990).

We removed all LLM integration directly from the CLI tool so that we do not have to deal with Gemini connections from a client. All Gemini integration now takes place within the Ditto backend, so that the CLI simply does the repository scan, uploads the results to S3 (via signed URL generated in the Ditto public API) and initiates the job to classify and organize all user-facing text.

## Test Plan

- [x] Make sure your CLI is pointed at local
- [x] Make sure you are also running the Ditto public API on the branch from [this ditto-app PR](https://github.com/dittowords/ditto-app/pull/8990)
- [x] From CLI terminal -- `yarn scan ~/Projects/ditto-pay-demo` (with your own test repo path)
- [x] You should get a Ditto app URL -- Copy the `/scan/:scanId` portion and go to that URL in the local app
- [x] You should be able to complete the A2 flow.


lib/src/scan/lang/i18n-file-discovery.ts
```diff
@@ -51,20 +36,149 @@ export const I18N_FILE_EXTENSIONS: ReadonlySet<string> = new Set([
...
+  "languages",
+  "strings",
+  "nls",
+  "intl",
+  "l10n",
+]);
+
+// BCP-47 locale token: 2-letter language code, optionally followed by
+// separator + 2-4 char region/script tag (e.g. en, fr, de-DE, zh_CN).
+const LOCALE_TOKEN_RE = /^[a-z]{2}([_-][a-zA-Z]{2,4})?$/i;
```
jholiga: i wonder if there's a way to match against the exact set so we don't match a db.json file or something similar

jholiga (approved):
testing went well! one small comment that might or might not be worth looking at

lib/src/scan/lang/i18n-file-discovery.ts
joustrich: Good call, just updated!

PR #140, opened 2026-06-11
[DIT-12985] Add locale_key and i18n_key to candidates

joustrich:
## Overview

Add locale_key and i18n_key to candidates

## Context

We use this when de-duplicating text across multiple languages and for variant creation

## Test Plan

- [x] Used in combination with this [ditto-app PR](https://github.com/dittowords/ditto-app/pull/9032). Follow the instructions there.


jholiga (approved):
Works great in the matrix repo!

PR #141, opened 2026-06-15
Ditto scan

joustrich:
## Overview

This PR introduces the `scan` command into the Ditto CLI. The scan command takes a single input, a path, and scans all files at that path to identify all user-facing text. The user-facing text candidates are then fed into Ditto where they can be turned into Ditto components from within the Ditto web application.

## Context

This is a base branch, so there have already been a number of PRs/reviews to get to this point. Here are the existing PRs that have been reviewed:
- https://github.com/dittowords/cli/pull/138
- https://github.com/dittowords/cli/pull/139
- https://github.com/dittowords/cli/pull/140

Only the [latest commit for some final cleanup](https://github.com/dittowords/cli/pull/141/changes/cbb297b9a787719a2e8f22c14518fe3c50def45d) is new and would be good to get reviewed more thoroughly.

## Test Plan

- [ ] Quick regression test of `pull`
- [ ] Run `scan` on a repository with user facing text ([matrix-poll](https://github.com/nordeck/matrix-poll) is a good example with both variants and plurals). You should be able to follow the link to complete the flow in the web app.


lib/src/formatters/mixins/javascriptCodegenMixin.ts
```diff
@@ -1,6 +1,6 @@
 import { Constructor } from "../shared";
 
-interface NamedImport {
+export interface NamedImport {
```
jholiga: is there a reason this is exported now? I can't remember

lib/src/utils/DittoError.ts
```diff
@@ -50,6 +50,8 @@ export default class DittoError<T extends ErrorType> extends Error {
 export enum ErrorType {
   ConfigYamlLoadError = "ConfigYamlLoadError",
   ConfigParseError = "ConfigParseError",
+  GeminiError = "GeminiError",
```
jholiga: Is this still needed?

lib/src/services/apiToken/promptForApiToken.ts
```diff
@@ -14,12 +14,11 @@ export const validate = async (token: string) => {
  * @returns The collected token
  */
 export default async function promptForApiToken() {
-  // @ts-expect-error - Enquirer types are not updated for the validate function
   const response = await prompt<{ token: string }>({
     type: "input",
     name: "token",
     message: "What is your API key?",
-    validate,
+    validate: validate as any,
```
jholiga, line 21: should this be included as part of this PR?

jholiga (approved):
Just a few small comments. Everything else looks good! Approving

lib/src/services/apiToken/promptForApiToken.ts
joustrich: I believe the tsconfig changes for exporting the types is what made this necessary.

lib/src/formatters/mixins/javascriptCodegenMixin.ts
joustrich: Oh dang apparently this is necessary. I had removed it and am getting errors attempting to run scan. Will add it back in.

PR #142, opened 2026-07-10
[DIT-12111] Improve error messages, particularly for errors from the Ditto API

joustrich:
## Overview

This PR improves error messages, particularly for errors from the Ditto API.

## Context

Previously any errors from the Ditto API would have logged out a generic error message, now it will instead log whatever error message is returned from the API.

## Screenshots

Before:
ditto-cli-project-error-before

After:
ditto-cli-project-error-after

## Test Plan

Testing successfully completed in dev/local via:

- [ ] Add an invalid project ID to your ditto/config.yml such as `id: "invalid-project-id"`
- [ ] Run the `pull` command (or you can use `yarn sync` if running directly from this repo)
- [ ] You should now see an error message that mentions you have an invalid project ID, like you can see in the "After" screenshot above. Previously this would have been the generic error you see in the "Before" screenshot above.

### DIFF
--- a/lib/src/index.ts
+++ b/lib/src/index.ts
@@ -13,7 +13,7 @@ import appContext from "./utils/appContext";
 import { ErrorType, isDittoError, isDittoErrorType } from "./utils/DittoError";
 import processCommandMetaFlag from "./utils/processCommandMetaFlag";
 
-const handleCommandError = async (error: unknown) => {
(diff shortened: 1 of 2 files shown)


marla-hoggard (approved):
Well that was easy! Hopefully all of our errors are caught and re-thrown correctly so thsi doesn't bleed any unwanted error message details to the user.

PR #143, opened 2026-07-21
[DIT-13305] Show error message when a scan exceeds plan limits. Make path param for scan optional.

jholiga:
## Overview
This PR introduces a couple small changes so that the CLI and specifically `ditto scan` are more productionized:
1. Show a formatted error message when a scan is run on a free or trial plan and exceeds the limit of extracted candidate strings (1000)
2. Make the path param for the scan optional. Omitting the path defaults to the cwd. This allows a user to run `npx @dittowords/cli scan` which will install, prompt the user for an API token, and run the scan from the current directory, all in one command.

## Context

Goes hand-in-hand with PR: [here
](https://github.com/dittowords/ditto-app/pull/9215)

## Screenshots

image

## Test Plan

- [ ] Ensure unit tests pass
- [ ] Test the CLI works on a repo without passing the path argument, e.g. `yarn scan` from the cli repo will run the scan on the CLI repo.
- [ ] Run the ditto scan on a repo that trips the max candidates error and ensure the error displays properly with a link to upgrade plan

### DIFF
--- a/lib/src/commands/scan.ts
+++ b/lib/src/commands/scan.ts
@@ -141,7 +141,7 @@ export const scan = async (
     const {
       candidatesSignedS3Url,
       record: { _id: recordId },
-    } = await initiateScan(path);
+    } = await initiateScan(resolvedInput);
     await uploadCandidatesToS3(candidates, candidatesSignedS3Url);
     await initiateClassify(recordId);
     logExtractSummary(extractSummary);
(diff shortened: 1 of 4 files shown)


joustrich (approved):

jholiga:
Going to merge but hold off on a full CLI package release, @tomkowalski will be doing a fast followup of this work that builds on the trial scan string limit functionality.

PR #145, opened 2026-07-27
[DIT-13328] Add better handling of limits for scan

tomkowalski:
## Overview
 Better handle cases where the `scan` command hits workspace limits around string classification. 

Upon initial scan the api returns the plan's limit, currently used budget, and plan name. This is used to help identify if the scan that is about to be done is possible within the workspace's limits. If it is not (or if it is run and we get back an error from the backend that a limit is reached) we will provide suggestions for directories that fit within the limit. 

The logic used for the directory suggestion is: 
* the directory's count of candidates, which is the sum of it's candidates and any of its sub directories, is less than the limit.
* the directory's parent directory is the root directory or another directory whose candidate count is > limit.
* sorted by candidate count desc for the matched directories, limited to a max of 10 suggestions.

There is also a `--list-directories` flag that will show the count of the candidates in every directories and its sub directories in lexicographical order to help with directory selection. 

Bump the cli version to reflect this change which is a minor update that does not influence the functionality, it just helps better shows the impact of the backend filtering.  

## Screenshots

Screenshot 2026-07-27 at 5 47 24 PM
Screenshot 2026-07-27 at 5 47 44 PM
Screenshot 2026-07-27 at 5 48 13 PM
Screenshot 2026-07-27 at 5 58 18 PM

## Test Plan

Testing successfully completed locally with:

- [x] Run ditto scan locally `DITTO_API_HOST=http://localhost:3001 DITTO_APP_HOST=http://localhost:3000 \
DITTO_TOKEN=$(curl -s "http://localhost:3001/trusted/token?email=<env>" | jq -r .token) \
yarn scan <path> --list-directories`
- [x] counts of the number of strings in each directory should be shown. 
- [x] Re-run with a small directory (count <1000) and no `--list-directories` flag in an non-limited org.
- [x] Follow the url to allow for the job to summarize.
- [x] Re-run with a small directory (count >1000) and no `--list-directories` flag in a limited org
- [x] There should be an error message with potential smaller directories to scan
- [x] Re-run with a small directory (count <1000) and no `--list-directories` flag in a limited org
- [x] Follow the url to allow for the job to summarize. Wait for it to finish
- [x] Re-run with a small directory (count <1000) and no `--list-directories` flag in a limited org
- [x] The new limit should be reflect based on the previous job's used limit.

### DIFF
--- a/lib/src/scan/analyzeDirectories.ts
+++ b/lib/src/scan/analyzeDirectories.ts
@@ -1,0 +1,184 @@
+import path from "path";
+
+// The rollup key for the input root. Candidate file paths are POSIX and
+// relative to the scanned root, so the root itself is the empty prefix.
+const ROOT = "";
+
+// The literal we suggest users run.
+const CLI_INVOCATION = "npx @dittowords/cli scan";
+
+export interface DirectorySuggestion {
+  dir: string;
+  count: number;
+}
+
+export interface DirectoryAnalysis {
+  total: number;
+  limit: number;
+  // Largest subtrees that each fit under the limit, sorted by descending count
+  // (tie-break: directory name ascending), capped.
+  suggestions: DirectorySuggestion[];
+  // Number of fitting subtrees beyond the cap.
+  omittedFittingCount: number;
+}
+
+export function analyzeDirectories(
+  filePaths: string[],
+  limit: number,
+  cap = 10
+): DirectoryAnalysis {
+  const total = filePaths.length;
+  const subtree = rollupCandidateCountsToParentPaths(filePaths);
+
+  // A directory fits (<= limit) and its parent does not.
+  // ROOT never qualifies because we only get here when total > limit.
+  const suggestedPaths: DirectorySuggestion[] = [];
+  for (const [dir, count] of subtree) {
+    if (dir === ROOT || count > limit) continue;
+    const parent = parentDirName(dir);
+    const parentCount = parent === null ? total : subtree.get(parent) ?? total;
+    if (parentCount > limit) suggestedPaths.push({ dir, count });
+  }
+
+  // Sort by descending count, tie-break by directory name; cap.
+  suggestedPaths.sort((a, b) => b.count - a.count || (a.dir < b.dir ? -1 : 1));
+  const cappedSuggestedPaths = suggestedPaths.slice(0, cap);
+  const omittedFittingCount =
+    suggestedPaths.length - cappedSuggestedPaths.length;
+  return {
+    total,
+    limit,
+    suggestions: cappedSuggestedPaths,
+    omittedFittingCount,
+  };
+}
+
+// Given a list of candidate file paths, returns a map of paths to the sum of candidates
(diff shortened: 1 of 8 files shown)


lib/src/scan/analyzeDirectories.ts
```diff
@@ -0,0 +1,184 @@
...
+  for (const path of filePaths) {
+    for (const dependentDir of parentDirsOf(path)) {
+      candidateCountByPath.set(
+        dependentDir,
+        (candidateCountByPath.get(dependentDir) ?? 0) + 1
+      );
+    }
+  }
+  return candidateCountByPath;
+}
```
tomkowalski, line 71: This is the core of the candidate counting. It counts each candidate once per directory in its path.

lib/src/scan/analyzeDirectories.ts
```diff
@@ -0,0 +1,184 @@
...
+  opts: { plan?: string; originalPath: string }
+): string {
+  const { total, limit, suggestions, omittedFittingCount } = analysis;
+  const root = displayRoot(opts.originalPath);
+
+  const header = `This scan found ${fmtCount(
+    total
+  )} strings, but your ${planLabel(opts.plan)} allows ${fmtCount(
+    limit
+  )} per scan.`;
```
jholiga: Since this is now a workspace limit, should we remove the "per scan", since it's cumulative?

lib/src/scan/analyzeDirectories.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  if (suggestions.length === 0) {
+    return [
+      header,
+      "",
+      `There's no subdirectory small enough to scan on its own — all ${fmtCount(
+        total
+      )} strings are in files at the top level of \`${root}\`. Point the scan at a narrower path to stay under the limit.`,
```
jholiga: I wonder if we should remove the "all strings are in files at the top level" messaging here, since this also catches the scenarios when there is a subdirectory but it's just too big.

lib/src/scan/analyzeDirectories.ts
```diff
@@ -0,0 +1,184 @@
...
+  const { total, limit, suggestions, omittedFittingCount } = analysis;
+  const root = displayRoot(opts.originalPath);
+
+  const header = `This scan found ${fmtCount(
+    total
+  )} strings, but your ${planLabel(opts.plan)} allows ${fmtCount(
+    limit
+  )} per scan.`;
+
+  if (suggestions.length === 0) {
```
jholiga, line 103: Do we want to have specific handling for the case where the budget has been completely used?

jholiga (commented):
Testing works well! I added a couple small comments around mostly wording of messages

lib/src/scan/analyzeDirectories.ts
tomkowalski: I think the chance of exactly hitting the limit is low and edge case-y enough that better language is probably unnecessary.

jholiga (approved):

PR #154, opened 2026-08-19
[DIT-13461]: Read git context in the CLI scan

laurakoye:
## Overview

New file `lib/src/scan/git.ts` — reads which repo and commit a scan came from, so a second scan can recognize a codebase Ditto has already seen.

```ts
readGitContext("/path/to/repo")
// {
//   repoKey: "github.com/dittowords/cli",   // from `origin`, same for every clone
//   repoRoot: "/Users/laura/…/cli",         // local only, never uploaded
//   commitSha: "d3c1a81…",
//   branch: "master",                       // null when HEAD is detached
//   dirty: false                            // uncommitted changes present
// }
```

`normalizeRepoKey` gives one stable key regardless of clone style, and credentials can't leak into it:

| remote URL | key |
|---|---|
| `git@github.com:Ditto/App.git` | `github.com/ditto/app` |
| `https://github.com/ditto/app.git` | `github.com/ditto/app` |
| `https://user:token@github.com/ditto/app.git` | `github.com/ditto/app` |
| `git@gitlab.com:group/subgroup/app.git` | `gitlab.com/group/subgroup/app` |

Keys are validated against `REPO_KEY_PATTERN`, copied from `ditto-app` (`shared/types/ProductTextDetection.ts:46`).

Nothing throws: no git, no repo, no commits, or no `origin` all return `null`. No new dependency — `node:child_process`.

**Nothing calls this yet.** #155 wires it into `POST /v2/scan`, so `ditto scan` is unchanged here.

## Context

- [DIT-13461](https://linear.app/dittowords/issue/DIT-13461/read-git-context-in-the-cli-scan). Blocks DIT-13467 (#155, stacked on this branch) and DIT-13468.
- Also bumps `@dittowords/text-extract` to `^0.3.0` so the scan picks up the shipped extractor work. The CLI imports only `runExtract` and two types, so nothing breaking reaches it — but candidate ids all change, since `makeCandidateId` now hashes `file:value:occurrence_index`.

## Test Plan

- [ ] `yarn test` passes — `git.test.ts` covers the URL forms above, a detached HEAD, a dirty tree, and the `null` cases; `extract.test.ts` covers the `0.3.0` bump
- [ ] `ditto scan` output unchanged from `master`
- [ ] Seeing the git context run end to end happens in #155

### DIFF
--- a/lib/src/scan/git.ts
+++ b/lib/src/scan/git.ts
@@ -3,17 +3,32 @@ import { promisify } from "node:util";
 
 const execFileAsync = promisify(execFile);
 
+/**
+ * Where a scan came from. `repoKey` and `commitSha` identify the same commit
+ * from any clone; `repoRoot` is local to this machine and isn't uploaded.
+ */
 export interface GitContext {
+  /** Host and path of `origin`, normalized: `github.com/dittowords/cli`. */
   repoKey: string;
+  /** Absolute path to the working tree. Local only — used to make candidate paths repo-root-relative. */
   repoRoot: string;
   commitSha: string;
+  /** `null` when HEAD is detached, as it is after `actions/checkout`. */
   branch: string | null;
+  /** Uncommitted changes are present, so `commitSha` doesn't fully describe what was scanned. */
   dirty: boolean;
 }
 
-const REPO_KEY_PATTERN =
+/**
+ * Copied from `shared/types/ProductTextDetection.ts` in `ditto-app`
+ */
+export const REPO_KEY_PATTERN =
   /^[a-z0-9.-]+(\/[a-z0-9._-]+){1,}\/(?!.*\.git$)[a-z0-9._-]+$/;
 
+/**
+ * The scp-like remote form `git@github.com:Ditto/App.git`, which has no scheme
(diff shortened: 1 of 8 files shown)


lib/src/scan/git.ts
```diff
@@ -0,0 +1,119 @@
...
+ * Copied from `shared/types/ProductTextDetection.ts` in `ditto-app`
+ */
+export const REPO_KEY_PATTERN =
+  /^[a-z0-9.-]+(\/[a-z0-9._-]+){1,}\/(?!.*\.git$)[a-z0-9._-]+$/;
+
+/**
+ * The scp-like remote form `git@github.com:Ditto/App.git`, which has no scheme
+ * and so isn't a URL. Captures the host and the path around the colon.
+ */
+const SCP_LIKE = /^(?:[^/@]+@)?([^/:]+):(.+)$/;
```
jholiga, line 32: Should we handle the edge case of if there's a port in the path, like `git@github.com:2222/ditto/app.git`?

jholiga (approved):
Just found one edge case with the regex for SCP, otherwise looks good!

jholiga:
Oh also don't forget to bump the version!

PR #155, opened 2026-08-19
[DIT-13467]: send git metadata with post v2scan

laurakoye:
## Overview

Stacked on #154 — this diff is only the wiring.

`ditto scan` now tells the server which repo and commit it scanned, so a later scan can be matched against this one instead of importing every string again.

What `POST /v2/scan` sends:

```diff
  {
    "path": "/Users/laura/Desktop/Ditto/cli/lib/src/scan",
+   "repoKey": "github.com/dittowords/cli",
+   "gitCommitSha": "d3c1a8148580e1869c91ee6caadda17165ceb0ea",
+   "gitBranch": "laura/dit-13467-send-git-metadata-with-post-v2scan",
+   "repoRelativeRoot": "lib/src/scan"
  }
```

`repoRelativeRoot` is where the scan started inside the repo. Candidate paths come out relative to
the scanned directory, so ditto-app #9367 prefixes this root to build a repo-root-relative code link
path. `""` when the scan is the repo root, and omitted if the scanned path somehow sits outside the
repo, so a bad value never becomes a wrong link.

Outside a git repo the body is unchanged from today, and the scan still runs:

```
[ditto scan] not a git repository - this scan can be imported but not re-synced
```

With uncommitted changes it runs too, with a note that the commit is approximate.

Safe to merge before the server reads any of this — zod strips unknown keys, so `/v2/scan` ignores the new fields today.

## Context

- [DIT-13467](https://linear.app/dittowords/issue/DIT-13467/send-git-metadata-with-post-v2scan). Blocked by DIT-13461 (#154).
- `repoRelativeRoot` is the first of [DIT-13468](https://linear.app/dittowords/issue/DIT-13468)'s fields, pulled forward so #9367 has its prefix. `scannedPaths` and `scannedAllPaths` stay in that ticket.
- `repoRoot` and `dirty` are read but not sent: `repoRoot` is local-only — it only derives `repoRelativeRoot` — and `dirty` is a warning, not a field.

## Test Plan

- [ ] `yarn test` passes — `buildInitiateScanBody` covers the payload with and without git context, at the repo root, and outside the repo
- [ ] `ditto scan` inside a repo completes as before
- [ ] `ditto scan` in a directory that isn't a repo prints the warning and still completes

### DIFF
--- a/lib/src/http/scan.ts
+++ b/lib/src/http/scan.ts
@@ -1,6 +1,7 @@
 import { DittoScanCandidate } from "@dittowords/text-extract";
 import axios, { AxiosError } from "axios";
 import { Blob } from "buffer";
+import { relative, sep } from "node:path";
 import { GitContext } from "../scan/git";
 import DittoError, { ErrorType } from "../utils/DittoError";
 import getHttpClient from "./client";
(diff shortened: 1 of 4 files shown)


lib/src/http/scan.ts
```diff
@@ -68,12 +73,32 @@ export function scanLimitError(
...
+/**
+ * Builds the `POST /v2/scan` body. Without git context the body is exactly what
+ * the CLI has always sent, so a scan outside a repo is unaffected.
+ */
+export function buildInitiateScanBody(
+  path: string,
+  gitContext?: GitContext | null
+): IInitiateScanBody {
+  if (!gitContext) return { path };
+  return {
```
jholiga, line 103: should we be passing the repo-relative root here?

jholiga (commented):
One comment, I think we want to be passing the repo-relative root in the post /v2/scan body, so we can use it for building the codelinks

jholiga (approved):

PR #157, opened 2026-08-27
ARB + ICU json frameworks | android + iOS locales output directories

joustrich:
## Overview

- New `arb` framework for the `json` format
- New `icu` framework for the `json` format (was previously just format `json_icu`, which still works but logs a warning and is marked deprecated)
- New `androidLocales` to match `iosLocales`
- New `androidLocalesOutDir` and `iosLocalesOutDir` for outputting android/iOS files to different root directories

## Context
See [ditto-app PR](https://github.com/dittowords/ditto-app/pull/9425) for corresponding backend and documentation changes

## Test Plan

Testing successfully completed in `local` via:

- [ ] Make sure you have a few variants set up with corresponding locales
- [ ] Make sure your config has the following (see my example config below)
  - [ ] `all` variants
  - [ ] `android`, `ios-strings`, `ios-stringsdict`, `json` (w/ framework `arb`) formats
  - [ ] `iosLocales` and `androidLocales` specified for all of your variants
  - [ ] `outDir`, `iosLocalesOutDir` and `androidLocalesOutDir` all specified to separate web/iOS/android outputs
- [ ] Once you have the config set up against your projects, run `pull` (or `yarn sync` if running from this project)
- [ ] You should see your ARB json files in the `outDir`, iOS files in the appropriate structure in your `iosLocalesOurDir` and android files in the appropriate structure in `androidLocalesOutDir`.

## Sample config

```
projects:
  - id: "jao-test"
variants:
  - id: "base"
  - id: "spanish"
  - id: "french"
  - id: "german"
outputs:
  - format: android
  - format: ios-strings
  - format: ios-stringsdict
  - format: json
    framework: arb
outDir: "ditto/json"
iosLocales:
  - base: en
  - spanish: es
  - french: fr
  - german: de
iosLocalesOutDir: "ditto/ios"
androidLocales:
  - base: en
  - spanish: es
  - french: fr
  - german: de
androidLocalesOutDir: "ditto/android"
```

### DIFF
--- a/lib/src/formatters/android.ts
+++ b/lib/src/formatters/android.ts
@@ -66,6 +66,21 @@ export default class AndroidXMLFormatter extends BaseExportFormatter<
       return `${localesOutDir}/values`;
     }
     const variantLocale = this.getVariantLocale(variantId);
-    return `${localesOutDir}/values-${variantLocale![variantId]}`;
+    const qualifier = this.toAndroidLocaleQualifier(variantLocale![variantId]);
+    return `${localesOutDir}/values-${qualifier}`;
+  }
+
+  /**
+   * Converts a locale code (e.g. "es-MX") into Android's resource-qualifier folder
+   * name (e.g. "es-rMX"). Android requires a 2-letter region subtag to be prefixed
+   * with a lowercase "r" — "values-es-MX" is not a valid qualifier and Android
+   * silently ignores the folder, so this must not be passed through as-is.
+   */
+  private toAndroidLocaleQualifier(locale: string): string {
+    const [language, region] = locale.split(/[-_]/);
+    if (region && /^[A-Za-z]{2}$/.test(region)) {
+      return `${language.toLowerCase()}-r${region.toUpperCase()}`;
+    }
+    return locale;
   }
 }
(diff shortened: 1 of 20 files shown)


lib/src/formatters/android.ts
```diff
@@ -9,16 +10,62 @@ export default class AndroidXMLFormatter extends BaseExportFormatter<
...
+      return this.outDir;
+    }
+    const localesOutDir =
+      this.projectConfig.androidLocalesOutDir ?? appContext.outDir;
+    const isBaseVariant = !variantId || variantId === BASE_VARIANT_ID;
+    if (isBaseVariant) {
+      return `${localesOutDir}/values`;
+    }
+    const variantLocale = this.getVariantLocale(variantId);
+    return `${localesOutDir}/values-${variantLocale![variantId]}`;
```
laurakoye: Region code break here. We should add a test to catch this case too. 

Android picks string files by folder name, abd a region has to be written with a lowercase r in from of it: 

values-es is spanish
values-es-rMX is mexican spanish
values-es-MX is not a valid folder name so Android ignores it

My local test shows that these are files that got generated. 

Screenshot 2026-08-27 at 6 13 12 PM

 Docs: ["The r prefix is used to distinguish the region portion."](https://developer.android.com/guide/topics/resources/providing-resources)

laurakoye (approved):
there is one regional bug for arb, but other than that it's good

lib/src/formatters/android.ts
joustrich: Good catch -- this is fixed!

PR #158, opened 2026-09-01
[DIT-13470] collect git file rename data

jholiga:
## Overview

CLI half of https://linear.app/dittowords/issue/DIT-13470/rename-pass. Before a scan starts, the CLI asks the server which commit the last scan read. It diffs that commit against `HEAD` with `git diff -M` and records what git says was moved or renamed.

Something to note is that git's similarity index decides what counts as a rename, not us. A heavily rewritten file is not a rename.

Every failure path returns no renames but never blocks the scan. That covers no git, no earlier scan, a shallow clone, an unreachable server, etc. A move of more than 1000 files is capped..

## Context

- [DIT-13470](https://linear.app/dittowords/issue/DIT-13470)
- ditto-app PR: https://github.com/dittowords/ditto-app/pull/9443

Merge the ditto-app PR first.

## Test Plan

- Ensure a ditto scan runs successfully. We can't test a re-scan until reconciliation is wired in.

### DIFF
--- a/lib/src/commands/scan.ts
+++ b/lib/src/commands/scan.ts
@@ -12,8 +12,10 @@ import {
 import chalk from "chalk";
 import {
   asScanLimitInfo,
+  getLastScanSha,
   initiateClassify,
   initiateScan,
+  MAX_SCAN_RENAMES,
   scanLimitError,
   uploadCandidatesToS3,
 } from "../http/scan";
@@ -22,7 +24,7 @@ import {
   formatDirectoryBreakdown,
   formatOverLimitMessage,
 } from "../scan/analyzeDirectories";
-import { readGitContext } from "../scan/git";
+import { GitContext, readGitContext, readRenames } from "../scan/git";
 import initAPIToken from "../services/apiToken/initAPIToken";
 import appContext from "../utils/appContext";
 import DittoError, { ErrorType } from "../utils/DittoError";
@@ -211,11 +213,13 @@ export const scan = async (
 
     const token = await initAPIToken();
     appContext.setAuthToken(token);
+    const renamesSinceLastScan = await readRenamesSinceLastScan(gitContext);
+
     const {
       candidatesSignedS3Url,
       record: { _id: recordId },
       planLimit,
-    } = await initiateScan(resolvedInput, gitContext);
+    } = await initiateScan(resolvedInput, gitContext, renamesSinceLastScan);
 
     // Fail before the wasted upload when the candidates we already extracted exceed it.
     if (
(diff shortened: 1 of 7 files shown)


lib/src/http/scan.ts
```diff
@@ -145,6 +158,22 @@ export async function initiateScan(
...
+ */
+export async function getLastScanSha(repoKey: string): Promise<string | null> {
+  try {
+    const httpClient = getHttpClient({});
+    const response = await httpClient.get("/v2/scan/last-commit", {
+      params: { repoKey },
+    });
+    return ZGetLastScanShaResponse.parse(response.data).lastScanSha ?? null;
+  } catch {
+    return null;
```
laurakoye, line 178: Should we be logging here?

lib/src/commands/scan.ts
```diff
@@ -269,3 +273,25 @@ export const scan = async (
     await quit(null, 0);
   }
 };
+
+/**
+ * What git says moved between the last scan of this repo and HEAD, so a renamed/moved file
+ * keeps its links instead of reading as a delete plus a create. Empty when there is
+ * no git context, no earlier scan, or no way to reach the earlier commit.
+ */
+async function readRenamesSinceLastScan(gitContext: GitContext | null) {
```
laurakoye: type the return

lib/src/scan/git.ts
```diff
@@ -117,3 +117,44 @@ export async function readGitContext(
...
+    if (status === "R" || status === "C") {
+      const [from, to] = [fields[i + 1], fields[i + 2]];
+      if (status === "R" && from && to) renames.push({ from, to });
+      i += 3;
+    } else {
+      i += 2;
+    }
+  }
+  return renames;
+}
```
laurakoye, line 161: I believe this can be simplified. Adding `--diff-filter=R` makes git send back only the renames. Every record is then the same shape, so the parser becomes a step by 3 loop.

Worth it mainly for the output size.

```suggestion
  const out = await git(
    ["diff", "-M","--diff-filter=R","--name-status","-z",previousSha,"HEAD"],
    repoRoot
  );
  if (!out) return [];

  const fields = out.split("\0").filter(Boolean);
  const renames: GitRename[] = [];
  for (let i = 0; i + 2 < fields.length; i += 3) {
    renames.push({ from: fields[i + 1], to: fields[i + 2] });
  }
  return renames;
```

laurakoye (approved):

lib/src/scan/git.ts
jholiga: nice catch!

lib/src/http/scan.ts
jholiga: going to skip the logging here, since this is a valid path for initial scans where no existing scan exists yet.

PR #159, opened 2026-09-14
[DIT-13659]: CLI: refuse a scan from a dirty tree or a non-default branch

laurakoye:
## Overview

`ditto scan` now refuses to run unless you're on a clean checkout of the default branch. It exits 1 with one of:

| Checkout | Message |
|---|---|
| Uncommitted changes | `Uncommitted changes are present. Commit or stash them, then scan again.` |
| Non-default branch | `On branch "my-feature", but this repo's default branch is "main". Check out "main" and scan again.` |
| Detached HEAD | `HEAD is detached at abc1234. Check out "main" and scan again.` |

The check runs before we ask for an API key, so a refused scan costs nothing — no login prompt, no API call, nothing against the plan limit.

**Working out the default branch.** We read `origin/HEAD`, falling back to a local `main` or `master`. That ref is recorded when you clone and never updates on its own, so a default branch renamed since would refuse a checkout that's actually correct — and name a branch that no longer exists. To avoid that, a mismatch is confirmed against the remote with `git ls-remote --symref origin HEAD` before we refuse. Successful scans never make that call; the round trip is only paid on the path that's about to stop you. If the remote is unreachable we fall back to the local ref and say so in the message.

Still scans as before:

- **Not a git repo.** Nothing to check, and these scans send no git info anyway. Same warning as today.
- **We can't work out the default branch at all.** A few repos have no `origin/HEAD` and no local `main` or `master`, usually one that was never cloned or whose default is named something like `develop`. Blocking there would refuse a checkout that's probably fine, with an error that can't say which branch to switch to.

## Context

A scan records the commit and branch it read, and code links point at them. A scan from a dirty tree or a feature branch points at lines the default branch never had, so the next scan reports those links as changed or removed, with no obvious cause.

Closes DIT-13659.

## Test Plan

Testing successfully completed locally via:

- [ ] `npx jest` — full suite green
- [ ] Clean checkout of the default branch scans as before
- [ ] Edit a file, then scan → exits 1, dirty message, no API key prompt
- [ ] Scan from a non-default branch → exits 1, names both branches
- [ ] `git checkout --detach HEAD`, then scan → exits 1, names the sha
- [ ] Scan a folder outside any git repo → still warns and scans


lib/src/scan/git.ts
```diff
@@ -159,3 +159,28 @@ export async function readRenames(
...
   return renames;
 }
+
+/**
+ * The branch a GitHub scan would read. `git clone` records it as `origin/HEAD`;
+ * a hand-built or CI checkout often has no such ref, hence the fallbacks.
+ *
+ * @returns `null` when none of the three resolve.
+ */
+export async function readDefaultBranch(
```
jholiga, line 177: super edge case, but in the case that the remote default branch is renamed, the locally cloned repo has no way of knowing / updating to that and would still only allow scans of the original default. I wonder if we want to add some sort of upstream check to handle that? something like `git ls-remote --symref origin HEAD` I think might do it?

jholiga (approved):

lib/src/scan/git.ts
laurakoye: Thanks! Added the check so we won't refuse on the rename case, as long as the remote is reachable.

